PR #18, opened 2018-10-24
Allow CLI transport to be selected and not require an access token

tommoor:
Local use should not require an access token.

Note: Because AUTO instantiates both transports it still acts as required there.


amccloud:
Ahh good point. The token is required for the collections cli endpoints, because that proxies back to api. Maybe a runtime assertion in this methods?

tommoor:
The CLI returns an error in that circumstance – do you think that would be okay to forward rather than reimplement error handling on the SDK side?

PR #24, opened 2018-10-26
Sketch Plugin Compatability

tommoor:
These changes allow the SDK to run in the environment of a Sketch plugin.

### DIFF
--- a/src/AbstractCLI/index.js
+++ b/src/AbstractCLI/index.js
@@ -102,8 +102,6 @@ export default class AbstractCLI implements AbstractInterface {
         resolve(data);
       });
 
-      abstractCli.stdout.setEncoding("utf8");
-
       abstractCli.stdout
         .on("data", data => {
           stream.write(data);
--- a/src/AbstractCLI/index.js
+++ b/src/AbstractCLI/index.js
@@ -63,6 +63,8 @@ export default class AbstractCLI implements AbstractInterface {
     this.accessToken = accessToken;
     this.apiUrl = apiUrl;
 
+    console.log(locatePath.sync(cliPath));
+
     try {
       this.cliPath = path.relative(
         cwd,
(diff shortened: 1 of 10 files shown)


src/AbstractCLI/index.js
```diff
@@ -93,21 +90,28 @@ export default class AbstractCLI implements AbstractInterface {
...
       });
 
+      const stream = JSONStream.parse();
+      stream.on("data", data => {
+        logStdoutData(data);
+        resolve(data);
+      });
+
       abstractCli.stdout
-        .pipe(JSONStream.parse())
```
tommoor: Pipe is not available in Sketch so we need to forward data events instead

src/AbstractCLI/index.js
```diff
@@ -80,7 +77,7 @@ export default class AbstractCLI implements AbstractInterface {
         : [];
 
       const spawnArgs = [
-        `./${path.relative(this.cwd, this.cliPath)}`,
+        this.cliPath,
```
tommoor: Relative path resolution is broken in Sketch

src/AbstractCLI/index.js
```diff
@@ -64,10 +64,7 @@ export default class AbstractCLI implements AbstractInterface {
     this.apiUrl = apiUrl;
 
     try {
-      this.cliPath = path.relative(
-        cwd,
-        path.resolve(cwd, locatePath.sync(cliPath))
-      );
+      this.cliPath = path.resolve(cwd, locatePath.sync(cliPath));
```
tommoor: This appeared to be doubling up on the logic below on L80, so also needed to go

src/AbstractCLI/index.js
amccloud: I originally did this to make snapshots easier but now paths are replaced with a placeholder

src/AbstractCLI/index.js
```diff
@@ -93,21 +90,28 @@ export default class AbstractCLI implements AbstractInterface {
       logSpawn(spawnArgs);
       const abstractCli = spawn(...spawnArgs);
 
-      let stderrBuffer = new Buffer.from("");
```
amccloud: good catch! We should probably rename `Abstract.Client` to `Abstract.client` to follow the same factory convention. I've also done `new Abstract.Client()` 🙈

PR #26, opened 2018-10-27
Add typescript declarations

gilad-segal:
Hey guys,

Following our design tools hackathon in Israel, here is my modest contribution to your project - typescript declarations.
There were a few unclear things about the documentation so I used `any` a few times.
Those are the uses:
1. Type `User` doesn't exist, it is used in: `Collections.user`, `Comment.user` and `Project.createdByUser`
2. `Collection.layers` is of type `CollectionLayer` which is doesn't exist
3. `LayerData.properties` is of type `LayerProperties` which doesn't exist

One more thing - I think you should consider creating types for `list` queries' filters. So signatures will look like this:
```
list: (
  descriptor: FileDescriptor | PageDescriptor,
  filter?: Partial<BranchListFilter>
) => Promise<Layer[]>;
```
Instead of: 
```
list: (
  descriptor: FileDescriptor | PageDescriptor,
  filter?: {limit?: number; offset?: number}
) => Promise<Layer[]>;
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,348 @@
+declare module 'abstract-sdk' {
+  namespace AbstractSdk {
+    export function Client(params: {
+      accessToken: string;
+      transport: TRANSPORTS;
+    }): AbstractClient;
+
+    export enum TRANSPORTS {
+      API,
+      CLI
+    }
+
+    export type AbstractClient = {
+      branches: {
+        list: (
+          description: ProjectDescriptor,
+          filter?: {filter?: 'active' | 'archived' | 'mine'}
+        ) => Promise<Branch[]>;
+
+        info: (descripiton: BranchDescriptor) => Promise<Branch>;
+      };
+
+      collections: {
+        list: (
+          descriptor: ProjectDescriptor | BranchDescriptor,
+          filter?: {layersPerCollection?: number}
+        ) => Promise<Collection[]>;
+
+        info: (descriptor: CollectionDescriptor) => Promise<Collection>;
+      };
+
+      comments: {
+        create: (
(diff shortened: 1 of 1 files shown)


tommoor:
Thanks Gilad, really great to have this included in the library 👍

amccloud:
@gilad-segal  🎉 thanks for the first contribution and for the tips! Option types are coming.

Re: typescript - It would be really nice to automate creating the typescript types from our flowtypes. My concern is that we don't use typescript internally, so typescript will unlikely be maintained.

Do you think the output of [flow-to-typescript](https://github.com/bcherny/flow-to-typescript#typescript-vs-flow) would work?

gilad-segal:
Perhaps you can compare the output of `flow-to-typescript` to this PR and see if it generates the same output. Personally, I wouldn't relay on it as many features are still missing and the project is in pre-alpha version.
There are a few concepts that exist only in typescript / flow, you can find out more [here](https://github.com/niieani/typescript-vs-flowtype) so I would prefer not to relay on a generator.
As I see it you have three options:
1. Maintain the typescript declarations in your project
2. Use a CI tool to generate it for you
3. Move this PR into [definitely typed](https://github.com/DefinitelyTyped/DefinitelyTyped) repo and relay on open source contributions to keep the typescript declarations in sync

@amccloud

bitpshr:
Hi @gilad-segal, thanks a ton for all of your effort on this. Unfortunately, the structure of the codebase changed significantly up until `abstract-sdk@1.0.0`, so this declaration file is no longer accurate. For now, I'm going to close this pull request, but please don't hesitate to reopen either this or a new pull request if you get around to creating a new TypeScript declaration file before we manage to do so.

Thanks again for to improve the SDK!

PR #27, opened 2018-10-29
Sketch Utility Methods

tommoor:
Note: This branch is based off `sketch-compat` – if the branch gets merged first the base needs changing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,6 @@
   "scripts": {
     "prepublish": "yarn run build",
     "build": "babel ./src -d ./lib",
-    "watch": "yarn build --watch",
     "lint": "eslint .",
     "test": "jest",
     "test-inspect": "node --inspect-brk ./node_modules/.bin/jest --runInBand",
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
   "scripts": {
     "prepublish": "yarn run build",
     "build": "babel ./src -d ./lib",
+    "watch": "yarn build --watch",
     "lint": "eslint .",
     "test": "jest",
     "test-inspect": "node --inspect-brk ./node_modules/.bin/jest --runInBand",
--- a/src/sketch.js
+++ b/src/sketch.js
@@ -1,37 +1,62 @@
 // @flow
 /* global process NSString PROSketchBootstrap */
+import type { ProjectDescriptor, BranchDescriptor, FileDescriptor } from ".";
 
 export function isSketchPlugin() {
   // New versions of skpm set process type to "sketch". For older
   // versions we can infer from the presence of ObjectiveC classes
-  // $FlowFixMe
-  return process.type === "sketch" || !!NSString;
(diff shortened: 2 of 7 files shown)


src/sketch.js
```diff
@@ -0,0 +1,37 @@
+// @flow
+/* global process NSString PROSketchBootstrap */
+
+export function isSketchPlugin() {
+  // New versions of skpm set process type to "sketch". For older
+  // versions we can infer from the presence of ObjectiveC classes
+  // $FlowFixMe
+  return process.type === "sketch" || !!NSString;
```
amccloud: What was flow complaining about here? Checking for undefined would be preferable over  `!!`

package.json
```diff
@@ -10,6 +10,7 @@
   "scripts": {
     "prepublish": "yarn run build",
     "build": "babel ./src -d ./lib",
+    "watch": "yarn build --watch",
```
amccloud: Probably no need to add this. `yarn build --watch` should work just like `yarn test --watch`

src/sketch.js
```diff
@@ -0,0 +1,37 @@
...
+  return documentKey(context).split("/")[0];
+}
+
+export function branchId(context: *) {
+  return documentKey(context).split("/")[1];
+}
+
+export function fileId(context: *) {
+  return documentKey(context).split("/")[3];
+}
```
amccloud: Could be interesting to make these return full descriptors `project(), branch(), file()`

tommoor: Yea, I'm into that.

package.json
tommoor: It'll work, but the shorthand is nice and makes it clear to folks that watch is an available option

amccloud: But what about yarn test —watch. Isn’t it nice that it’s symetrical and works out of the box?

src/sketch.js
tommoor: It was complaining on that it couldn't be found, I think I need to make flow understand it's a global but I wrote this on the plane and didn't have access to flow docs. Will try and remove the FixMe's here.

src/sketch.js
```diff
@@ -1,37 +1,62 @@
...
+  };
+}
+
+export function file(context: *): FileDescriptor {
+  const key = documentKey(context);
+
+  return {
+    projectId: projectId(key),
+    branchId: branchId(key),
+    fileId: fileId(key)
```
amccloud: Nit: split gets called 3 times here. Maybe:

```js
const [projectId, branchId, _, fileId] = documentKey(context).split("/"); // idk what _ is but maybe name that too?
```

tommoor: True, but I'd prefer to keep the knowledge of where the individual id's are positioned in the key in one place rather than copied in each method – I think the slight performance hit is worth the DRYness.

Ideally we'd move away from splitting the document key in the future, it's kind of a hack anyway.

PR #28, opened 2018-10-29
Fill out draft of contributing section

tommoor:
Anything else we should add?


amccloud:
@tommoor could mention something about docs and test

PR #29, opened 2018-10-30
Shares

amccloud:
⚠️ Depends on https://github.com/goabstract/projects/pull/823

- Add `ShareDescriptor` from url or id
- Add `shares.info`
- Add `descriptor.info`


amccloud:
👍 that would be easier. Something like `share.descriptor`?

tommoor:
If you wanted to make it backwards compatible that would work – I was thinking literally change the share JSON response as the only real difference seems to be `commitSha -> sha` ?

amccloud:
I think we might send additional properties too. It would be nice to match the exact type of the descriptor without having to omit or anything.

docs/reference.md
```diff
@@ -657,10 +719,16 @@ Reference for the parameters required to load resources with Abstract SDK.
 
 ```js
 {
-  projectId: string,
+  projectId: string
 }
 ```
 
+### ShareDescriptor
```
bitpshr: Nit: Should we also document composite types like `ShareableDescriptor`?

docs/reference.md
```diff
@@ -645,8 +646,69 @@ abstract.projects.list({
...
+
+or, by id…
+
+```js
+abstract.shares.info({
+  id: '49b1f582-a8b4-46ca-8c86-bbc675fe27c4'
+});
+```
+
+### Convert share link into descriptor
```
bitpshr: This markdown page tries to provide both API documentation and general reference guide documentation, which can be a little confusing. Here we talk about converting share URLs into descriptors that can be used with other SDK calls under the section for `abstract.shares`, but this doesn't involve `abstract.shares` at all.

Not urgent or specific to this PR, but I wonder if there would be value in providing (generated) API documentation separately from curated reference material.

src/AbstractAPI/index.js
```diff
@@ -144,13 +147,46 @@ export default class AbstractAPI implements AbstractInterface {
...
     list: async () => {
       const response = await this.fetch("organizations");
       return unwrapEnvelope(response.json());
     }
   };
 
+  shares = {
+    info: async (shareUrl: string) => {
+      const response = await this.fetch(
+        `share_links/${parseShareURL(shareUrl)}`
```
bitpshr: The added shares documentation states that shares can be queried by ID, but it seems like they can only be queried by URL. Maybe `parseShareURL` should normalize any URL or ID into one or the other, and then both this and `descriptors` can use that common function. This would also get rid of the double-ternary in `descriptors`, which is a little hard to read.

docs/reference.md
amccloud: I think maybe when it becomes public api

src/AbstractAPI/index.js
amccloud: Updated this to take `ShareDescrptor` which seems to align really well with everything else

docs/reference.md
amccloud: Removed `descriptors.info` in favor of `share.descriptor`

docs/reference.md
amccloud: `ShareableDescriptor` is now removed

docs/reference.md
```diff
@@ -645,8 +646,69 @@ abstract.projects.list({
...
+A share is a shareable link to an object. Use the desktop or web app to create share links.
+
+### The share object
+
+| Property         | Type     | Description                                                                                    |
+|------------------|----------|------------------------------------------------------------------------------------------------|
+| `descriptor`     | [Descriptor](#Descriptor) | A descriptor for the shared object                                            |
+| `url`            | `string`                  | The share url                                                                 |
+| `color`          | `string`                  | A hex value that represents a custom project color                            |
+| `createdAt`      | `string`                  | Timestamp that the project was created                                        |
```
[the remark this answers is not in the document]
amccloud: Updated to reflect the properties in flow

amccloud:
@bitpshr thanks for the review. This is ready for another pass when you get a chance

src/index.js.flow
```diff
@@ -60,6 +62,8 @@ export type LayerDescriptor = {|
   layerId: string
 |};
 
+export type ShareDescriptor = {| url: string |} | {| shareId: string |};
```
amccloud: Also renamed this to shareId

src/support/factories.js
```diff
@@ -113,6 +113,7 @@ export function buildCommentDescriptor(
   commentDescriptor: *
 ): CommentDescriptor {
   return {
+    projectId: "project-id",
```
amccloud: Not related to this branch but flow was complaining. Might have been a bad merge but `projectId` is required for comment now. Not necessarily a bad thing either

amccloud: > must be an artifact of a bad merge somewhere

Yea I had two `CommentDescriptors`, one with and without a project id and I  wasn't sure which one to take

> Part of me worries about maintainability if our SDK becomes increasingly asymmetrical with the backing API.

This is part of the goal of the SDK. We realized that the API had rough edges and is not ready to consumed directly. I think it's okay to introduce asymmetry here if we're rounding off problems with the underlying API with the intent to go fix it after

Not saying this is a problem now but I suspect including `projectId` will allow for better read performance as the comment table grows and we [partition it by something like project](https://github.com/goabstract/projects/blob/master/stable/api/db/structure.sql#L1546). Looking at the metrics `comments#show` is never used, so the sdk will likely be the first real consumer of this api

tommoor: >  I suspect including projectId will allow for better read performance

🙏

amccloud:
Removed `projectId` from CommentDescriptor and will move that into another PR

Issue #21, opened 2018-10-25
Add activity endpoints


tommoor:
- [x] activities.list (see project, org and more filters available in API controller)
- [ ] activities.info

PR #33, opened 2018-11-12
API: Initial support for activity lists

bitpshr:
This pull request adds support for `activities.list` to the API transport.

Refs #21

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -6,6 +6,36 @@ title: Reference
 [cli-icon]: https://img.shields.io/badge/CLI-lightgrey.svg
 [api-icon]: https://img.shields.io/badge/API-blue.svg
 
+## Activities
+
+![API][api-icon]
+
+An activity represents a designated type of event within a project. These events can be speceific to the project itself, or they can be specific to a collection, a branch, a commit, or a review within the project.
+
+### The activity object
+
+| Property      | Type                | Description                                                                             |
+|---------------|---------------------|-----------------------------------------------------------------------------------------|
+| `branchId`    | `string`            | UUID of the branch that this activity was triggered on                                  |
+| `createdAt`   | `string`            | Timestamp at which the activity was triggered                                           |
+| `id`          | `string`            | UUID identifier of the activity                                                         |
+| `payload`     | `Object`            | Object containing information specific to this type of activity                         |
+| `type`        | `string`            | The type of this activity, may be one of `BRANCH_ARCHIVED`, `BRANCH_CREATED`, `BRANCH_DELETED`, `BRANCH_DESCRIPTION_UPDATED`, `BRANCH_RENAMED`, `BRANCH_STATUS_UPDATED`, `BRANCH_UNARCHIVED`, `COLLECTION_PUBLISHED`, `COMMENT_CREATED`, `COMMIT`, `MERGE_COMMIT`, `PROJECT_ARCHIVED`, `PROJECT_CREATED`, `PROJECT_DELETED`, `PROJECT_DESCRIPTION_CHANGED`, `PROJECT_RENAMED`, `PROJECT_TRANSFERRED`, `PROJECT_UNARCHIVED`, `REVIEWER_REMOVED`, `REVIEW_COMPLETED`, `REVIEW_DISMISSED`, `REVIEW_REQUESTED`, `UPDATE_COMMIT` |
+| `userId`      | `string`            | UUID of the user that triggered this activity                                           |
+
+### List all activities
+
+`activities.list(BranchDescriptor | OrganizationDescriptor | ProjectDescriptor, { limit?: number, offset?: number }): Promise<Activity[]>`
+
+List all activities for a given project on a specific branch
+
+```js
+abstract.activities.list({
+  branchId: "8a13eb62-a42f-435f-b3a3-39af939ad31b",
+  projectId: "b8bf5540-6e1e-11e6-8526-2d315b6ef48f",
+}, { limit: 2 });
+```
+
 ## Branches
 
 ![API][api-icon]
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -10,7 +10,7 @@ title: Reference
 
 ![API][api-icon]
 
-An activity represents a designated type of event within a project. These events can be speceific to the project itself, or they can be specific to a collection, a branch, a commit, or a review within the project.
+An activity represents a designated type of event within a project. These events can be specific to the project itself, or they can be specific to a collection, a branch, a commit, or a review within the project.
 
 ### The activity object
 
@@ -27,7 +27,7 @@ An activity represents a designated type of event within a project. These events
 
 `activities.list(BranchDescriptor | OrganizationDescriptor | ProjectDescriptor, { limit?: number, offset?: number }): Promise<Activity[]>`
 
-List all activities for a given project on a specific branch
+List the first two activities for a given project on a specific branch
 
 ```js
 abstract.activities.list({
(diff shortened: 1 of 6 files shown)


src/index.js.flow
```diff
@@ -59,6 +59,44 @@ export type Cursor<T> = Promise<{
   // [Symbol.asyncIterator](): string
 };
 
+export type FilterOptions = {
```
bitpshr: I figured this type could also be used in intersection types for various calls as pagination is added throughout, e.g.:

```js
export type BranchListOptions = ListOptions & { filter?: "active" | "archived" | "mine" };
```

src/index.js.flow
```diff
@@ -59,6 +59,44 @@ export type Cursor<T> = Promise<{
   // [Symbol.asyncIterator](): string
 };
 
+export type ListOptions = {
```
bitpshr: I figured this type could also be used in intersection types for various calls as pagination is added throughout, e.g.:

```js
export type BranchListOptions = ListOptions & { filter?: "active" | "archived" | "mine" };
```

src/index.js.flow
```diff
@@ -59,6 +59,44 @@ export type Cursor<T> = Promise<{
...
+  limit?: number,
+  offset?: number
+};
+
+export type Activity = {
+  branchId: string,
+  createdAt: string,
+  id: string,
+  payload: Object,
```
amccloud: Would it be crazy/unnecessary to type out payload for each type? 🙊

src/AbstractAPI/index.js
```diff
@@ -212,17 +228,6 @@ export default class AbstractAPI implements AbstractInterface {
...
-    info: async (branchDescriptor: BranchDescriptor) => {
-      const response = await this.fetch(
-        // prettier-ignore
-        `projects/${branchDescriptor.projectId}/branches/${branchDescriptor.branchId}`
-      );
-
-      return response.json();
-    }
-  };
-
```
amccloud: 🤔

src/AbstractAPI/index.js
```diff
@@ -138,6 +139,21 @@ export default class AbstractAPI implements AbstractInterface {
...
 
+  activities = {
+    list: async (
+      objectDescriptor?:
+        | BranchDescriptor
+        | OrganizationDescriptor
+        | ProjectDescriptor,
+      options: ListOptions = {}
+    ) => {
+      const query = queryString.stringify({ ...options, ...objectDescriptor });
```
amccloud: I think we should be a bit more defensive here an explicitly map the options and descriptor to query like https://github.com/goabstract/abstract-sdk/pull/33/files#diff-54ed2903b1e3820fa4ffcf86da9efc93R340

package.json
```diff
@@ -11,6 +11,7 @@
     "prepublish": "yarn run build",
     "build": "babel ./src -d ./lib",
     "lint": "eslint .",
+    "lint:fix": "eslint . --fix",
```
amccloud: This shouldn't be necessary. `yarn lint --fix` should work out of the box 👍

src/AbstractAPI/index.js
bitpshr: I was confused by this one too at first. It looks like two `branch` members were included unintentionally, one of which only implemented `info`.

docs/reference.md
```diff
@@ -6,6 +6,36 @@ title: Reference
...
+| `id`          | `string`            | UUID identifier of the activity                                                         |
+| `payload`     | `Object`            | Object containing information specific to this type of activity                         |
+| `type`        | `string`            | The type of this activity, may be one of `BRANCH_ARCHIVED`, `BRANCH_CREATED`, `BRANCH_DELETED`, `BRANCH_DESCRIPTION_UPDATED`, `BRANCH_RENAMED`, `BRANCH_STATUS_UPDATED`, `BRANCH_UNARCHIVED`, `COLLECTION_PUBLISHED`, `COMMENT_CREATED`, `COMMIT`, `MERGE_COMMIT`, `PROJECT_ARCHIVED`, `PROJECT_CREATED`, `PROJECT_DELETED`, `PROJECT_DESCRIPTION_CHANGED`, `PROJECT_RENAMED`, `PROJECT_TRANSFERRED`, `PROJECT_UNARCHIVED`, `REVIEWER_REMOVED`, `REVIEW_COMPLETED`, `REVIEW_DISMISSED`, `REVIEW_REQUESTED`, `UPDATE_COMMIT` |
+| `userId`      | `string`            | UUID of the user that triggered this activity                                           |
+
+### List all activities
+
+`activities.list(BranchDescriptor | OrganizationDescriptor | ProjectDescriptor, { limit?: number, offset?: number }): Promise<Activity[]>`
+
+List all activities for a given project on a specific branch
```
tommoor: "List all activities" -> "List the first 2 activities"

docs/reference.md
```diff
@@ -6,6 +6,36 @@ title: Reference
 [cli-icon]: https://img.shields.io/badge/CLI-lightgrey.svg
 [api-icon]: https://img.shields.io/badge/API-blue.svg
 
+## Activities
+
+![API][api-icon]
+
+An activity represents a designated type of event within a project. These events can be speceific to the project itself, or they can be specific to a collection, a branch, a commit, or a review within the project.
```
tommoor: Spelling mistake here "speceific", suggest:

> An activity is an event in Abstract, for example creating a commit, renaming a branch, or inviting a new member to a project. They offer a good way to get a feed of what's happening for an organization, project, or branch.

package.json
tommoor: In the case of Visual Studio Code you can set `eslint.autoFixOnSave=true` in the config to have the editor do this, not sure which editor is your preferred Paul?

src/index.js.flow
tommoor: Reference: https://github.com/goabstract/ui-core/blob/master/src/components/Activity/types.js

src/AbstractAPI/index.js
bitpshr: Ah, I naively thought these descriptor types each only had a single property so spreading would be safe, but I see now the `BranchDescriptor` contains additional properties we don't want to add to the query string. Great catch. I think spreading `options` should be OK though since the `ListOptions` interface only defines properties we want to include.

src/index.js.flow
bitpshr: Thanks a ton for that link, huge time saver 🙏

bitpshr:
Thanks for the quick review @amccloud and @tommoor. I believe all comments have been addressed.

src/AbstractAPI/index.js
amccloud: I mainly suggested being defensive because not everyone using the SDK will be using flow. Also, `ListOptions` would need to be an exact type (if it's not already) to guarantee we don't pass extra properties.

src/index.js.flow
```diff
@@ -64,11 +64,252 @@ export type ListOptions = {
...
+  | PayloadUpdateCommit;
+
 export type Activity = {
   branchId: string,
   createdAt: string,
   id: string,
-  payload: Object,
+  payload: Payload,
   projectId: string,
   type:
```
amccloud: @bitpshr is the type in the payload too? I was sorta expecting:

```js
export type Activity = ActivityBranchArchived | ActivityBranchCreated | ...
```

I think right now it's typed as `payload.type` and `payload.payload`

src/AbstractAPI/index.js
bitpshr: Good info about the exact types, this is something I definitely need to remember coming from TypeScript.

amccloud: Hopefully flow will be changing the default to exact types pretty soon 🤞

docs/reference.md
tommoor: ⚡️  Just wanted to make sure you spotted this documentation suggested change.

PR #36, opened 2018-11-14
API: comments.list and comments.info

### LINKED ISSUE #30: Add comments endpoints
- [x] comments.create
- [ ] comments.list
- [ ] comments.info

bitpshr:
This pull request adds support for `comments.list` and `comments.info` to the API transport.

Resolves #30

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -200,8 +200,11 @@ or, list the first two comments for a specific layer...
 
 ```js
 abstract.comments.list({
-  layerId: '9FB27CF4-81F5-40A3-AB34-B2E5C792BEE3',
-  projectId: '616daa90-1736-11e8-b8b0-8d1fec7aef78'
+  branchId: "master",
+  fileId: "51DE7CD1-ECDC-473C-B30E-62AE913743B7",
+  layerId: "CA420E64-08D0-4B96-B0F7-75AA316B6A19",
+  pageId: "7D2D2599-9B3F-49BC-9F86-9D9D532F143A",
+  projectId: "616daa90-1736-11e8-b8b0-8d1fec7aef78"
 }, { limit: 2 });
 ```
 
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -187,11 +187,35 @@ represents a bounding area ontop of the layer, this can be used to leave comment
 
 ### List all comments
 
-  > Not yet implemented
+`comments.list(BranchDescriptor | CommitDescriptor | PageDescriptor | LayerDescriptor): Promise<Comment[]>`
+
+List the comments for a specific project
+```js
+abstract.comments.list({
+  projectId: "616daa90-1736-11e8-b8b0-8d1fec7aef78"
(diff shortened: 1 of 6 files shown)


docs/reference.md
```diff
@@ -175,11 +175,37 @@ represents a bounding area ontop of the layer, this can be used to leave comment
...
+  projectId: "616daa90-1736-11e8-b8b0-8d1fec7aef78"
+});
+```
+
+or, list the first two comments for a specific file on a specific branch...
+
+```js
+abstract.comments.list({
+  branchId: "master",
+  fileId: "51DE7CD1-ECDC-473C-B30E-62AE913743B7",
```
tommoor: I feel this isn't the greatest example because we actually don't allow commenting directly on a file, although it will probably work and aggregate comments on layers in the file.

I'd give an example of requesting comments for a layer instead

src/AbstractAPI/index.js
```diff
@@ -229,6 +230,28 @@ export default class AbstractAPI implements AbstractInterface {
...
+      options: ListOptions = {}
+    ) => {
+      const query = queryString.stringify({
+        limit: options.offset,
+        offset: options.offset,
+        branchId: objectDescriptor.branchId,
+        commitSha: objectDescriptor.sha,
+        fileId: objectDescriptor.fileId,
+        pageId: objectDescriptor.pageId,
+        projectId: objectDescriptor.projectId
```
tommoor: Shouldn't `layerId` be here somewhere

bitpshr:
Thanks for the review @tommoor. PR updated per your feedback and rebased against master.

docs/reference.md
```diff
@@ -187,11 +187,35 @@ represents a bounding area ontop of the layer, this can be used to leave comment
...
+abstract.comments.list({
+  projectId: "616daa90-1736-11e8-b8b0-8d1fec7aef78"
+});
+```
+
+or, list the first two comments for a specific layer...
+
+```js
+abstract.comments.list({
+  layerId: '9FB27CF4-81F5-40A3-AB34-B2E5C792BEE3',
```
tommoor: I think this example also needs to be a full layer descriptor… can copy the one from layers.info

bitpshr: It looks like the API resiliently handles requests that only provide a `projectId` and a `layerId`. Should we update our `LayerDescriptor` type accordingly, or keep it as-is if this is an invalid way to call the endpoint?

tommoor: @amccloud thoughts? Should we say to use a layerdescriptor even if you only _really_ need the `layerId` ? I'm on the fence tbh

tommoor: > layerId is not unique in a project.

True, yea – i'd hazard to say we might need to make the API error with a layerId but without the other parameters in the long run. It's probably a bug that it returns anything

bitpshr: Thanks for the context @amccloud and @tommoor.

PR #38, opened 2018-11-17
API: Cursor-based pagination

### LINKED ISSUE #32: Add pagination / cursor support
Currently list endpoints only return the first page of results

bitpshr:
This pull request adds support for cursor-based pagination to the API.

Resolves #32

### DIFF
--- a/docs/pagination.md
+++ b/docs/pagination.md
@@ -3,4 +3,47 @@ id: pagination
 title: Pagination
 ---
 
-Coming soon...
+Some SDK methods return large result sets that can be difficult to work with because of how many items they contain. To make working with such large sets easier and more efficient, these methods return a few items at a time in smaller sets called "pages".
+
+This section explains how to use these pages of results within the SDK.
+
+## Cursor pagination
+
+Instead of returning all results at once, paginated SDK methods return a **cursor** object. A cursor object contains a small portion of results - or a page - along with a function that can be called to fetch additional pages. Instead of manually managing options like `offset`, a cursor can be used to iteratively and automatically move through pages of results.
+
+In practice, paginated methods return a special type of `Promise` that contains a `next` method. This special `Promise` - called a `CursorPromise` - resolves to the current page of results. Its `next` method returns another `CursorPromise` that resolves to the next page of results, or `undefined` if none exist.
+
+## Example
+
+The following example demonstrates how to work with pages of results.
+
+1. Use an SDK client instance to fetch the first ten activities for a project.
+
+  ```js
+  const page = abstract.activities.list({
+    projectId: 'b8bf5540-6e1e-11e6-8526-2d315b6ef48f'
+  }, { limit: 10 });
+  ```
+
+2. Since `page` is a `CursorPromise`, it will resolve to the first ten activities. Its `next` method can be used to fetch the next page of activities.
+
+  ```js
+  console.log(`Result page: ${(await page).length} items`);
+  const nextPage = response.next();
+  console.log(`Result page: ${(await nextPage).length} items`);
+  ```
+
+3. Putting this all together, all activities could be fetched using recursion.
+
+  ```js
+  async function fetchActivities(request) {
+      const response = await request;
+      if (!response) return;
+      console.log(`Result page: ${response.length} items`);
+      fetchActivities(request.next());
+  }
+
+  fetchActivities(abstract.activities.list({
+    projectId: 'b8bf5540-6e1e-11e6-8526-2d315b6ef48f'
+  }, { limit: 10 }));
+  ```
--- a/src/AbstractAPI/index.js
+++ b/src/AbstractAPI/index.js
@@ -22,10 +22,13 @@ import type {
   CommentDescriptor,
   Comment,
   Layer,
-  ListOptions
+  ListOptions,
+  Activity,
+  Notification
 } from "../";
 import parseShareURL from "./parseShareURL";
 import randomTraceId from "./randomTraceId";
+import Cursor from "./Cursor";
 
 const minorVersion = version.split(".", 2).join(".");
 const logStatusError = log.extend("AbstractAPI:status:error");
@@ -157,22 +160,32 @@ export default class AbstractAPI implements AbstractInterface {
   }
 
   activities = {
-    list: async (
+    list: (
       objectDescriptor: $Shape<
         BranchDescriptor & OrganizationDescriptor & ProjectDescriptor
       > = {},
       options: ListOptions = {}
     ) => {
-      const query = queryString.stringify({
-        limit: options.offset,
-        offset: options.offset,
-        branchId: objectDescriptor.branchId,
-        organizationId: objectDescriptor.organizationId,
-        projectId: objectDescriptor.projectId
-      });
-      const response = await this.fetch(`activities?${query}`);
-      const { activities } = await unwrapEnvelope(response.json());
-      return activities;
+      return new Cursor<Activity[]>(
+        async (meta = { nextOffset: options.offset }) => {
+          const query = queryString.stringify({
+            limit: options.limit,
+            offset: meta.nextOffset,
+            branchId: objectDescriptor.branchId,
+            organizationId: objectDescriptor.organizationId,
+            projectId: objectDescriptor.projectId
+          });
+          const response = await this.fetch(`activities?${query}`);
+          const {
+            data: { activities },
+            meta: newMeta
+          } = await response.json();
+          return {
+            data: activities,
+            meta: newMeta
+          };
+        }
+      ).promise;
     },
     info: async ({ activityId }: ActivityDescriptor) => {
       const response = await this.fetch(`activities/${activityId}`);
@@ -279,7 +292,7 @@ export default class AbstractAPI implements AbstractInterface {
 
       return response.json();
     },
-    list: async (
+    list: (
       objectDescriptor: {
         projectId: $PropertyType<ProjectDescriptor, "projectId">
       } & $Shape<
@@ -287,19 +300,22 @@ export default class AbstractAPI implements AbstractInterface {
       >,
       options: ListOptions = {}
     ) => {
-      const query = queryString.stringify({
-        limit: options.offset,
-        offset: options.offset,
-        branchId: objectDescriptor.branchId,
-        commitSha: objectDescriptor.sha,
-        fileId: objectDescriptor.fileId,
-        layerId: objectDescriptor.layerId,
-        pageId: objectDescriptor.pageId,
-        projectId: objectDescriptor.projectId
-      });
-      const response = await this.fetch(`comments?${query}`);
-      const comments = await unwrapEnvelope(response.json());
-      return comments;
+      return new Cursor<Comment[]>(
+        async (meta = { nextOffset: options.offset }) => {
+          const query = queryString.stringify({
+            limit: options.limit,
+            offset: meta.nextOffset,
+            branchId: objectDescriptor.branchId,
+            commitSha: objectDescriptor.sha,
+            fileId: objectDescriptor.fileId,
+            layerId: objectDescriptor.layerId,
+            pageId: objectDescriptor.pageId,
+            projectId: objectDescriptor.projectId
+          });
+          const response = await this.fetch(`comments?${query}`);
+          return await response.json();
+        }
+      ).promise;
     },
     info: async ({ commentId }: CommentDescriptor) => {
       const response = await this.fetch(`comments/${commentId}`);
@@ -535,18 +551,21 @@ export default class AbstractAPI implements AbstractInterface {
   };
 
   notifications = {
-    list: async (
+    list: (
       objectDescriptor?: OrganizationDescriptor,
       options: ListOptions = {}
     ) => {
-      const query = queryString.stringify({
-        limit: options.offset,
-        offset: options.offset,
-        organizationId: objectDescriptor && objectDescriptor.organizationId
-      });
-      const response = await this.fetch(`notifications?${query}`);
-      const notifications = await unwrapEnvelope(response.json());
-      return notifications;
+      return new Cursor<Notification[]>(
+        async (meta = { nextOffset: options.offset }) => {
+          const query = queryString.stringify({
+            limit: options.offset,
+            offset: meta.nextOffset,
+            organizationId: objectDescriptor && objectDescriptor.organizationId
+          });
+          const response = await this.fetch(`notifications?${query}`);
+          return await response.json();
+        }
+      ).promise;
     },
     info: async ({ notificationId }: NotificationDescriptor) => {
       const response = await this.fetch(`notifications/${notificationId}`);
--- a/src/AbstractAPI/Cursor.js
+++ b/src/AbstractAPI/Cursor.js
@@ -1,0 +1,47 @@
+/* @flow */
+import type { CursorMeta, CursorPromise, CursorResponse } from "../";
+
+export default class Cursor<T> {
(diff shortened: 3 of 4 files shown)


docs/pagination.md
```diff
@@ -0,0 +1,81 @@
...
+  }, { limit: 10 });
+  ```
+
+3. Since `response` is a `Cursor` object, it will contain a `value` property with the first ten activities, and if more activities exist, it will also contain a `next` property. Using this, the next page of activities can be fetched.
+
+  ```js
+  console.log(`Result page: ${response.value.length} items`);
+  if (response.next) {
+    const nextResponse = await response.next();
+    console.log(`Result page: ${nextResponse.value.length} items`);
```
amccloud: We wouldn't want to change the semantics of our return type just because it's a cursor. Instead of `cursor.value`, have you tried awaiting the promise as usual to get the value?

```js
const cursor = abstract.activies.list();
console.log('Page 1', await cursor); // => [Activity, Activity, ...]
console.log('Page 2', await cursor.next()); // => [Activity, Activity, ...]
console.log('Page 3', await cursor.next().next()); // => [Activity]
```

amccloud:
@bitpshr gotcha. Async iteration was a stretch goal. Now that you mention it, I had planned on doing something like this for the recommended api / fallback for those who don't have async iteration:

```js
// Fallback or recommended API
Abstract.paginator({
 currentPage: 5 // Resume from page 5 of pagination
})(abstract.layers.list(FileDescriptor), (cursor, done) => {
  if (cursor.meta.currentPage > 10) done();
  const layers = await cursor;
});
```

That being said if changing the semantics to Promise<{value}>` to align with async iteration could be worth it... cc: @tommoor

tommoor:
😬 Feels like always having to get `.value` off the response is kind of annoying, particularly in the single item response scenario.

I'd vote we either drop async iteration for now considering that most devs probably aren't aware of it, or we have value on list responses only...?

amccloud:
I'm 👍 for dropping async iterators. I think a wrapper function like `Abstract.paginator` can add support if it returned the right signature

docs/pagination.md
```diff
@@ -0,0 +1,81 @@
+---
+id: pagination
+title: Pagination
+---
+
+Some SDK methods return large result sets that can be difficult to work with because of how many items they contain. To make working with such large collections easier and more efficient, these methods return a few items at a time in smaller collections called "pages".
```
amccloud: Maybe replace `collections` with `list` to avoid confusion with our collections?

docs/pagination.md
```diff
@@ -0,0 +1,81 @@
...
+
+This section explains how to use these pages of results within the SDK.
+
+## Cursor pagination
+
+Instead of returning all results at once, paginated SDK methods return small portions - or pages - of results along with a function that can be called to fetch the next page. This function is commonly referred to as a **cursor**. Instead of manually managing options like `offset`, a cursor can be called to iteratively move through pages of results.
```
amccloud: `a cursor can be called` -> `next() can be called on the cursor`

docs/pagination.md
```diff
@@ -0,0 +1,81 @@
...
+
+In practice, paginated methods return a `Promise` that resolves to a special type of object:
+
+```js
+type Cursor<T> = Promise<{
+  meta: {
```
amccloud: Tempted to call this `page` to align with pagination/pages etc. `page.total` vs `meta.total`, page seems to have more signal

docs/pagination.md
```diff
@@ -0,0 +1,81 @@
...
+
+2. Next, use the client to fetch the first ten activities for a project.
+
+  ```js
+  const response = abstract.activities.list({
+    projectId: 'b8bf5540-6e1e-11e6-8526-2d315b6ef48f'
+  }, { limit: 10 });
+  ```
+
+3. Since `response` is a `Cursor` object, it will contain a `value` property with the first ten activities, and if more activities exist, it will also contain a `next` property. Using this, the next page of activities can be fetched.
```
amccloud: 👀

bitpshr:
Hi @amccloud and @tommoor, thanks again for the review and the feedback so far.

I think native asynchronous iteration support is a powerful feature for a data-retrieving API. It also felt odd to put off support mainly due to API asymmetry. To that end, I updated this implementation so that all paginated methods conditionally return an asynchronously-iterable object if a new `asyncIterator` list option is passed. Logic paths were intentionally kept distinct: we can easily drop support for cursor pagination in the future as asynchronous iteration becomes more widely used.

Using cursor pagination:

```js
const response = await abstract.activities.list({ projectId: /* ... */});
console.log(response);
console.log(await response.next());
// ...
```

Using asynchronous iteration:

```js
const response = abstract.activities.list({ projectId: /* ... */}, {
    asyncIterator: true
});

for await (const value of response) {
    console.log(value);
}
```

All other feedback so far should be addressed as well.

amccloud:
@bitpshr thanks for pushing on this! I'd really like to not add an option to the interface of all list methods. Can we explore with our own usage first before adding it to the sdk? One question I think still need to be answered is will we iterate over the cursors or the items inside of the cursor?

I think we can hold off and use a wrapper function to handle async iteration for now:

```js
// untested
function paginate(cursor) {
  return {
    value: cursor,
    done: cursor.meta.nextOffset === undefined,
    [Symbol.asyncIterator]() {
      return {
        next() { return paginate(cursor.next()); }
      };
    }
  };
}

const paignatedActivities = paginate(abstract.activities.list({ projectId }));

for await (const cursor of paignatedActivities) {
  console.log(await cursor);
}
```

I think with this route we will have the flexiblity to switch between page/item itereration without touching the list methods

bitpshr:
Thanks for the feedback @amccloud.

I’m not sure I totally understand the rationale for providing a convenience method to handle asynchronous iteration instead of native support within each list method. From a high-level, your pseudocode implementation is more or less identical to the implementation in this PR, minus the type annotations. The main difference between these two approaches is usability: one requires users to import and use a standalone method to wrap cursor results instead of passing an option and getting an actual `AsyncIterator` interface. Both of these approaches require explicit opt-in from the user, and both approaches still require that we maintain the same underlying logic. The only tangible difference I see is that the three paginated list methods we support have four extra lines of logic. In my mind, this is a very acceptable tradeoff (if even considered a tradeoff at all) to provide result sets that are asynchronously iterable without being wrapped in another function. It should also be noted that no pagination logic is duplicated or maintained in any list method; instead, they just conditionally create the correct type of result set based on their options.

Candidly, I’m struggling to hone in on the actual reason not to support both types of result sets: Is it maintainability? Is it type complexity? No reason I can think of seems strong enough to force users to go through the burden of importing a separate function and wrapping results when our API can do it for them.

FWIW, and perhaps most importantly, if the single new option seems unmaintainable, we can ditch it altogether if we're OK with awaiting the initial result before asynchronous iteration:

```js
const response = await abstract.activities.list({ projectId });

console.log(response);
console.log(await response.next());

for await (const value of response) {
    console.log(value);
}
```

This means we can support both cursor pagination and asynchronous iteration out of the box with no wrapping function and no new list option. In this case, what's the argument for not supporting it?

amccloud:
> I’m not sure I totally understand the rationale for providing a convenience method to handle asynchronous iteration instead of native support within each list method.

Primarily maintainability. We have quite a bit we want to add to the SDK and trying to manage the surface area for complexity

> Re: iterating over individual items vs. pages of items, I think we have to iterate over pages of items to stay symmetrical with cursor pagination. Since we can't know how downstream users leverage the SDK

☝️Another reason to make the paginate a function vs. including it in the transports. Think of it as [inversion of control](https://en.wikipedia.org/wiki/Inversion_of_control). Pulling the logic out into a wrapper function will allow us (and 3rd parties) to do custom async iteration by page or item

We could do both but re: my first point on maintainability, I'm not convinced it is worth it _yet_

The paginate function could also handle things for you like resuming pagination, limiting the iteration, throttling. We can use functional composition to experiment with all of that without adding more options to the transports

All that being said your last example with no option is interesting! I'd probably still explore a paginate function around `response` if I wanted to change the iteration semantics

bitpshr:
Hi @amccloud, thanks for the context around externalizing asynchronous iteration.

It's odd to me to invert control for asynchronous iteration but to not do the same for cursor pagination. The advantages described above for control inversion apply to any pagination approach, not just asynchronous iteration. The described use case of paginating individual items also seems contrived: we don't offer that ability for cursor pagination, and individual items within each page are already iterable synchronously.

I still don't fully understand the rationale for intentionally unbalancing an API by externalizing one pagination method but not another. There's no additional maintenance burden since logic is isolated in a class outside of each list method (just like logic would be isolated in an external pagination function we'd provide.) Still, for the sake of landing cursor support independently of iteration support, I updated this PR to only include the former and created #41.

This should be ready for another review @amccloud, thanks very much in advance.

amccloud:
@bitpshr no problem! I realize we're pretty much on the same page and just want to provide some clarification on my part:

> It's odd to me to invert control for asynchronous iteration but to not do the same for cursor pagination. 

I want to make `Abstract.paginator` the recommended interface to paginate and consume cursors. I believe we still have to return a type for any paginator to consume - as our typical `Promise<Array<*>>` expose enough information to support pagination on their own. Open to other suggestions if you have another way to do this

In hindsight I didn't make that clear enough and focused too much on the stretch goal of supporting async iteration in my initial review

> I still don't fully understand the rationale for intentionally unbalancing an API by externalizing one pagination method but not another. There's no additional maintenance burden since logic is isolated in a class outside of each list method (just like logic would be isolated in an external pagination function we'd provide.)

Not my intentions, my intention is to add as few things as necessary from the get-go while we learn how the sdk is adopted. `Cursor` seems like the minimium API to  expose pagination possibilites without changing the original semantics of of the sdk while limiting our maintenance overhead

If we have a strong idea of what `Abstract.paginator` should be we should include it in the sdk. I do think if we include it would be nice for it to have async support. But moving forward I think `paginate(transport())` will be easier to maintain over `transport({paginate})`

bitpshr:
Sounds good @amccloud, thanks for the additional context. It seems like the best path forward is to implement a `paginator` function as part of #41, and to let this pull request serve as the base cursor implementation. I think this should still be ready for another review when you have a chance.

src/AbstractAPI/index.js
```diff
@@ -158,21 +160,26 @@ export default class AbstractAPI implements AbstractInterface {
...
+        projectId: descriptor.projectId
+      };
+      const response = await this.fetch(
+        `activities?${queryString.stringify(query)}`
+      );
+      const responseData = await response.json();
+      const nextPage = offset =>
+        this.activities.list(descriptor, { ...options, offset });
+
+      return new CursorResultSet(responseData, nextPage, "activities");
```
amccloud: Could be nice to update the body of cursor to accept the original code so that we can infer `nextPage` and not need to know about `valueProp`:

```js

list = (descriptor, options) => {
  return new Cursor(async cursor => {
    // original list code with access to cursor.meta
  });
}
```

*Note: I removed async on the list method to avoid wrapping the cursor in another promise*

src/AbstractAPI/CursorResultSet.js
```diff
@@ -0,0 +1,33 @@
...
+ * Array-like wrapper for cursor-paginated data
+ */
+export default class CursorResultSet<T> extends Array<T> {
+  /**
+   * Creates a new CursorResultSet
+   *
+   * @param response - Top-level JSON data from a list endpoint
+   * @param nextPage - Called to fetch the next page of results
+   * @param [valueProp] - Used to target nested response data
+   */
```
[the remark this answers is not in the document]
bitpshr: See #39. We still need a way to generate API documentation that's accurate. We can rely on Flow for typings, but we need subjective descriptions of parameters and members as well. https://documentation.js.org/ supports stripped-down JSDoc syntax, meaning it infers typings from Flow so only certain JSDoc information needs to be included. This is looking like the best solution so far.

src/AbstractAPI/CursorResultSet.js
```diff
@@ -0,0 +1,33 @@
+// @flow
+
+/**
+ * Array-like wrapper for cursor-paginated data
+ */
+export default class CursorResultSet<T> extends Array<T> {
```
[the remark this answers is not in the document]
bitpshr: No, there's no difference. We can change the name to `Cursor` if we intend to use the vernacular where the retrieval function is not the cursor, but the result set itself is.

docs/pagination.md
```diff
@@ -0,0 +1,51 @@
...
+title: Pagination
+---
+
+Some SDK methods return large result sets that can be difficult to work with because of how many items they contain. To make working with such large sets easier and more efficient, these methods return a few items at a time in smaller sets called "pages".
+
+This section explains how to use these pages of results within the SDK.
+
+## Cursor pagination
+
+Instead of returning all results at once, paginated SDK methods return small portions - or pages - of results along with a function that can be called to fetch the next page. This function is commonly referred to as a **cursor**. Instead of manually managing options like `offset`, a cursor can be used to iteratively move through pages of results.
```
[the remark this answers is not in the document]
bitpshr: Sounds good re: verbiage distinction, it makes sense to me to call the entire augmented `Array` a "cursor" in this case. I'm also totally on-board with nixing `next` (and other cursor-specific) documentation once we land a pagination function.

src/AbstractAPI/CursorResultSet.js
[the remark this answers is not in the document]
bitpshr: On second thought, I removed comments for now since we're not even sure what the correct syntax will end up being. We can handle comments here (and in all other modules) as part of #39.

src/AbstractAPI/CursorResultSet.js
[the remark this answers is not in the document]
amccloud: Just noticed `CursorResultSet` was an array. Thoughts on extending Promise, overriding construction and adding `next()`? To align with Cursor behaving just like a promise but can be paginated?

src/AbstractAPI/index.js
bitpshr: Great idea; this comment led me to a more recursive approach that allowed us to both ditch `valueProp` and maintain one logic path for query construction.

src/AbstractAPI/CursorResultSet.js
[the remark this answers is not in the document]
bitpshr: Check out the latest implementation in 22a76c32f4db16f637448efbe3850ca6b3586aa7 - I think we can get away with only extending `Array`. Also, from a high level, I'm not sure it makes sense to extend `Promise` in this case: the result set that the returned `Promise` resolves to should contain the `next` function, not the `Promise` itself:

```js
const items = await abstract.activities.list({ /* ... */ });
console.log(items);
const moreItems = await items.next();
console.log(moreItems);
// ...
```

bitpshr:
Thanks a ton @amccloud, I think the implementation is much nicer now that we ditched `valueProp`. Ready for another pass when you have a chance.

tommoor:
The API and docs we ended up with here are clear, I haven't followed all of the conversation above but was able to understand how to use this with ease.

tommoor:
What's it going to take to merge this this week gents?

bitpshr:
@tommoor I'm cleaning this up based on the approach in [this gist](https://gist.github.com/bitpshr/12a94f6aa0b210bba18ff7898e0cdb3c) per Andrew's suggestion (augmenting `Promise` instead of `Array`), just finishing up types now.

bitpshr:
The latest approach to cursor pagination fundamentally different, so for review clarity, I opened up a new PR. Closing in favor of #49.

PR #46, opened 2018-12-06
feature: document changesets.info

### LINKED ISSUE #31: Document changesets endpoint
`changesets.info` is implemented, but undocumented.

bitpshr:
This pull request adds documentation for `changesets.info`.

Resolves #31

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -106,6 +106,38 @@ abstract.branches.info({
 ```
 
 
+## Changesets
+
+![CLI][cli-icon] ![API][api-icon]
+
+A changeset is a group of changes that together form a single, indivisible modification to a repository. A changeset can be thought of as the differences between two successive commits in a repository's history. Changesets are automatically generated as new commits are created, and provide insight into what changed between two versions of a repository.
+
+### The changeset object
+
+| Property       | Type                | Description                                                                             |
+|----------------|---------------------|-----------------------------------------------------------------------------------------|
+| `branchId`     | `string`            | UUID of the branch that this changeset is part of, or the string "master"               |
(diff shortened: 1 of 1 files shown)


docs/reference.md
```diff
@@ -106,6 +106,38 @@ abstract.branches.info({
 ```
 
 
+## Changesets
+
+
+A changeset is a group of changes that together form a single, indivisible modification to a repository. A changeset can be thought of as the differences between two successive commits in a repository's history. Changesets are automatically generated as new commits are created, and provide insight into what changed between two versions of a repository.
```
tommoor: Please replace occurrences of "repository" with "project" as our preferred nomenclature

tommoor: Second sentence here seems to pretty much repeat the first, I think we can lose it? Changesets aren't generated, they are more a view of the data that's available.

docs/reference.md
```diff
@@ -106,6 +106,38 @@ abstract.branches.info({
...
+
+### The changeset object
+
+| Property       | Type                | Description                                                                             |
+|----------------|---------------------|-----------------------------------------------------------------------------------------|
+| `branchId`     | `string`            | UUID of the branch that this changeset is part of, or the string "master"               |
+| `changes`      | `ChangesetChange[]` | List of changes that make up this changeset                                             |
```
tommoor: We're going to have to define `ChangesetChange` but happy to land this PR ahead of that

bitpshr: Ah, good catch, updated.

It looks like we have a lot of other undocumented supporting types throughout. The tediousness of maintaining these by hand (and potential for inaccuracies since we're not deriving them from existing types) is a strong argument for generated API docs.

Issue #44, opened 2018-12-04
Add a "Retrieve the latest preview" section to the docs


tommoor:
https://github.com/goabstract/abstract-sdk/issues/40#issuecomment-443562588

Issue #40, opened 2018-11-25
API: preview.info


igorsheg-wix:
Hey, 

I'm trying to get the preview to work - to get an png of a layer.

I'm using `preview.info`

Getting `https://previews.goabstract.com/projects/eda68300-e1d0-11e8-8133-0b70a3eae7f0/commits/51bf03f17a6234d6f703b86d2309fe5ae4a28034/files/D06C3330-664C-4F89-B8E6-AC27B93AF4DC/layers/4DF3A97D-3F66-48A1-9459-B42B3154391C`

then when i try to access it I get `{"error":"forbidden"}`

maybe the documentation is unclear or I got something wrong.

Thanks in advance!

tommoor:
Hey Igor, short answer is – use `previews.raw` to get the data and save it somewhere. The previews service requires authentication with a token which would have to be saved with the URL. 

We're looking into how this can be improved (perhaps with a previews specific short-lived token) as we realize it's not as developer friendly as it should be.

tommoor:
> you should also be able to pass sha: "latest" in the descriptor

Ah, of course 🙈 – should have mentioned that too.

igorsheg-wix:
@tommoor @amccloud Thanks guys ;)

PR #48, opened 2018-12-07
feature: tweak documentation structure

bitpshr:
This pull request updates the structure of the documentation site in a few ways:

- The "Getting Started" page was expanded into a "Getting Started" section since there's a lot of content for one page now (especially if more client configuration parameters are ever added.)
- `Abstract.Sketch` API docs were moved into the "Reference" section alongside `Abstract.Client` docs so all API docs are in one place.
- A new "Concepts" section was added for things like pagination (#32), requesting latest commits ([#40](https://github.com/goabstract/abstract-sdk/issues/40#issuecomment-443562588)), and other concepts we identify as important-enough to document independently.

Resolves https://github.com/goabstract/abstract-sdk/issues/44

_Overview of modified structure_

installation

_New "Request Latest Commit" page_

request-latest


docs/authentication.md
```diff
@@ -7,6 +7,6 @@ Authentication with the SDK is achieved by using an OAuth access token. During t
 
   > API tokens should be treated like a password, as they provide access to all data that your user has access to. Do not commit them to version control.
 
-## Usage
+## Using a token
 
-To use the token simply save it in your environment as `ABSTRACT_TOKEN` and it will automatically be used. You can also use another method to load the token and pass it to the SDK on initialization – see [Getting Started](/docs/getting-started) for more details.
+To use the token simply save it in your environment as `ABSTRACT_TOKEN` and it will automatically be used. You can also use another method to load the token and pass it to the SDK on initialization – see [Configuration](/docs/configurationi) for more details.
```
tommoor: The link is bad here "configurationi"

tommoor:
Looks great in general, my only concern with this change is that having the bulk of the API documentation kind of hidden underneath "Abstract.Client" isn't very clear to me…

docs/configuration.md
```diff
@@ -0,0 +1,38 @@
+---
+id: configuration
+title: Configuration
+---
+
+In order to use the SDK you'll need to instantiate an instance and give it an [authentication](/docs/authentication) token. We recommend storing the token in your environment as `ABSTRACT_TOKEN` and it will be automatically loaded – don't commit it with your code!
```
amccloud: `need to instantiate a client`

docs/configuration.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Transports
+
+If you want to ensure that the SDK only ever loads data from the API or the CLI then you can achieve this by specifying a transport. This is useful when running in an environment without the Mac application installed or alternatively when you want to ensure you're only dealing with local data:
+
+```js
+import * as Abstract from "abstract-sdk";
+
+const abstract = Abstract.Client({
+  transport: Abstract.TRANSPORTS.CLI
```
amccloud: suggestion:

```
  transport: Abstract.TRANSPORTS.CLI // or Abstract.TRANSPORTS.API
```

docs/latest.md
```diff
@@ -0,0 +1,16 @@
...
+---
+
+When using the SDK to query for various types of resources, it's common to pass in an `sha` property to limit results to a specific commit on a given project branch. In addition to specifying an actual SHA hash, the latest commit can always be targeted by passing the string `"latest"`.
+
+For example, it's possible to fetch activities for the latest commit on the master branch of a project without knowing the actual commit hash:
+
+```js
+abstract.activities.list({
+  branchId: "master",
+  commitId: "latest",
```
amccloud: sha

docs/latest.md
```diff
@@ -0,0 +1,16 @@
+---
+id: latest
+title: Request latest commit
+---
+
+When using the SDK to query for various types of resources, it's common to pass in an `sha` property to limit results to a specific commit on a given project branch. In addition to specifying an actual SHA hash, the latest commit can always be targeted by passing the string `"latest"`.
+
+For example, it's possible to fetch activities for the latest commit on the master branch of a project without knowing the actual commit hash:
```
amccloud: I'm not sure latest will actually work on activities. As of now `latest` as the sha is only supported on `LayerDescriptor` and `FileDescriptor` and can only be used where those descriptors can be used

bitpshr:
Thanks for the feedback so far @amccloud and @tommoor, this should be ready for another pass.

Re: the main documentation being buried under `Abstract.Client`, I'm torn: I agree that it's not very obvious, but it's accurate per the structure of the API, which I do think is important if we're labeling it as actual API documentation. Maybe we can ditch the `Abstract.` package namespace in each menu item since it's sort of redundant, and move `Sketch` to a `util` namespace?

amccloud:
@bitpshr 👍 on removing the prefix. Could maybe even have Client and Sketch top-level

tommoor:
> move Sketch to a util namespace

Could you say more about this?

bitpshr:
@tommoor thinking ahead, will we have additional utility modules beyond `Sketch`? If so, it may make sense to nest those modules in a `util` namespace for clarity and organization:

```js
import {
    Client,
    util
} from "abstract-sdk";

const abstract = Client();

util.Sketch.isSketchPlugin();
util.Illustrator.someMethod();
```

Totally secondary to downstream use, this would also alleviate the documentation clarity issue in this PR since the API menu would have `Client` at the top level with `Sketch` nested under `util`, so we could one of the following:

nested

Or...

inline

tommoor:
Gotcha. I'd vote to keep it top level for now, if we end up with a top-level namespace for each tool we support that's not necessarily a bad thing – util / lib folders tend to become a dumping ground pretty fast 😅

bitpshr:
Ok, I ditched the `Abstract.` namespace on each menu item and left `Sketch` top level for now. This should be ready for another pass @tommoor and @amccloud.

Playing devil's advocate re: `util` outside of this PR (though I'm definitely not tied to it): without a `util` folder (or something more purposeful, like `vendor`) to place `Sketch` and other product-specific utilities, the root of the SDK package itself becomes the dumping ground. I've never felt that the potential dumping ground argument was strong reason to avoid clear package structures / module hierarchy. In either case, thought should always be put into the public APIs we expose regardless of what level we expose them at. I do think it's a nonissue for now though.

tommoor:
Sounds good, argument heard. Lets keep an eye on it as the SDK grows in scope and keep iterating!

PR #49, opened 2018-12-17
feature: add cursor-based pagination support

### LINKED ISSUE #32: Add pagination / cursor support

bitpshr:
This pull request adds support for cursor-based pagination.

From a high level, a paginated list method should return a special `Promise` that contains a `next` method. The most obvious way to satisfy this requirement is to extend the built-in `Promise` class. Unfortunately, Flow doesn't support class method overloads, but does support interface method overloads; this means it's seemingly impossible to subclass Flow's own `Promise` type interface since it includes [an overloaded `then` definition](https://github.com/facebook/flow/blob/master/lib/core.js#L618-L626) that can't actually be implemented with type safety.

To overcome this limitation, a combination of interface inheritance and indirect type casting was used. A `CursorPromise` interface was created that implements the `Promise` interface:

```js
interface CursorPromise<T> extends Promise<T> {
    next(): CursorPromise<T>;
}
```

An underlying `Cursor` class holds the actual implementation, and returned `Cursor` values are indirectly cast to `CursorPromise` objects (which is safe in this case since we know APIs are compatible):

```js
class Cursor<T> {
    next(): CursorPromise<T> {
       const response = /* ... */
       return ((response: any): CusorPromise<T>;
    }
}
```

Despite the type complexities, this implementation fully satisfies the API described in the [abstract-js doc](https://paper.dropbox.com/doc/abstract-js--ATFcyDMuLHE9NP36WHK7HhxIAg-sQIIqdls6ZRFpPsVABtK4#:uid=794781938248041307487637&h2=Pagination) in a type-safe manner for consumers:

```js
const firstPage = activities.list<Activity[]>();
const firstPageResults = await firstPage;
console.log(firstPageResults[0].foo); // Error: foo does not exist on Activity
console.log(firstPageResults[0].id); // 89f5e25b...
console.log(await firstPage); // [...]

const secondPage = firstPage.next();
console.log(await secondPage); // [...]

const thirdPage = secondPage.next();
console.log(await thirdPage) // undefined
```

Resolves #32

### DIFF
--- a/docs/pagination.md
+++ b/docs/pagination.md
@@ -20,17 +20,24 @@ The following example demonstrates how to work with pages of results.
 1. Use an SDK client instance to fetch the first ten activities for a project.
 
   ```js
-  const page = abstract.activities.list({
+  const firstPage = abstract.activities.list({
     projectId: 'b8bf5540-6e1e-11e6-8526-2d315b6ef48f'
   }, { limit: 10 });
   ```
 
-2. Since `page` is a `CursorPromise`, it will resolve to the first ten activities. Its `next` method can be used to fetch the next page of activities.
+2. Since `firstPage` is a `CursorPromise`, it will resolve to the first ten activities. Its `next` method can be used to fetch the next page of activities.
 
   ```js
-  console.log(`Result page: ${(await page).length} items`);
-  const nextPage = response.next();
-  console.log(`Result page: ${(await nextPage).length} items`);
+  const firstPage = abstract.activities.list({
+    projectId: 'b8bf5540-6e1e-11e6-8526-2d315b6ef48f'
+  }, { limit: 10 });
+
+  const firstPageResponse = await firstPage;
+
+  const secondPage = firstPage.next();
+  const secondPageResponse = await secondPage;
+
+  console.log(`First Page: ${firstPageResponse.length}, Second Page: ${secondPageResponse.length}`);
   ```
 
 3. Putting this all together, all activities could be fetched using recursion.
--- a/docs/pagination.md
+++ b/docs/pagination.md
@@ -3,4 +3,47 @@ id: pagination
 title: Pagination
 ---
 
-Coming soon...
+Some SDK methods return large result sets that can be difficult to work with because of how many items they contain. To make working with such large sets easier and more efficient, these methods return a few items at a time in smaller sets called "pages".
+
+This section explains how to use these pages of results within the SDK.
+
+## Cursor pagination
+
+Instead of returning all results at once, paginated SDK methods return a **cursor** object. A cursor object contains a small portion of results - or a page - along with a function that can be called to fetch additional pages. Instead of manually managing options like `offset`, a cursor can be used to iteratively and automatically move through pages of results.
+
+In practice, paginated methods return a special type of `Promise` that contains a `next` method. This special `Promise` - called a `CursorPromise` - resolves to the current page of results. Its `next` method returns another `CursorPromise` that resolves to the next page of results, or `undefined` if none exist.
+
+## Example
+
+The following example demonstrates how to work with pages of results.
+
+1. Use an SDK client instance to fetch the first ten activities for a project.
+
+  ```js
+  const page = abstract.activities.list({
+    projectId: 'b8bf5540-6e1e-11e6-8526-2d315b6ef48f'
+  }, { limit: 10 });
+  ```
+
+2. Since `page` is a `CursorPromise`, it will resolve to the first ten activities. Its `next` method can be used to fetch the next page of activities.
+
+  ```js
+  console.log(`Result page: ${(await page).length} items`);
+  const nextPage = response.next();
+  console.log(`Result page: ${(await nextPage).length} items`);
+  ```
+
+3. Putting this all together, all activities could be fetched using recursion.
+
+  ```js
+  async function fetchActivities(request) {
+      const response = await request;
+      if (!response) return;
+      console.log(`Result page: ${response.length} items`);
+      fetchActivities(request.next());
+  }
+
+  fetchActivities(abstract.activities.list({
+    projectId: 'b8bf5540-6e1e-11e6-8526-2d315b6ef48f'
+  }, { limit: 10 }));
+  ```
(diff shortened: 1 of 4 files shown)


docs/pagination.md
```diff
@@ -3,4 +3,47 @@ id: pagination
...
-Coming soon...
+Some SDK methods return large result sets that can be difficult to work with because of how many items they contain. To make working with such large sets easier and more efficient, these methods return a few items at a time in smaller sets called "pages".
+
+This section explains how to use these pages of results within the SDK.
+
+## Cursor pagination
+
+Instead of returning all results at once, paginated SDK methods return a **cursor** object. A cursor object contains a small portion of results - or a page - along with a function that can be called to fetch additional pages. Instead of manually managing options like `offset`, a cursor can be used to iteratively and automatically move through pages of results.
+
+In practice, paginated methods return a special type of `Promise` that contains a `next` method. This special `Promise` - called a `CursorPromise` - resolves to the current page of results. Its `next` method returns another `CursorPromise` that resolves to the next page of results, or `undefined` if none exist.
```
bitpshr: I know we've mentioned this before, but just for posterity: most of this documentation will change when we add a `paginate` function soon after this lands.

src/AbstractAPI/Cursor.js
```diff
@@ -0,0 +1,47 @@
+/* @flow */
+import type { CursorMeta, CursorPromise, CursorResponse } from "../";
+
+export default class Cursor<T> {
+  lastResponse: ?Promise<CursorResponse<T> | void>;
+  meta: CursorMeta;
+  request: (meta?: CursorMeta) => Promise<CursorResponse<T>>;
+
+  get promise(): CursorPromise<T> {
```
bitpshr: This is a convenience method that indirectly casts this `Cursor` to a `CursorPromise` since we know the former does actually implement the latter.

src/AbstractAPI/Cursor.js
```diff
@@ -0,0 +1,47 @@
...
+    onFulfilled: (data: T) => *,
+    onRejected: (error: *) => *
+  ): CursorPromise<T> {
+    const cursorPromise = this._next(true);
+    cursorPromise.then(onFulfilled, onRejected);
+    return cursorPromise;
+  }
+
+  next(first?: boolean): CursorPromise<T> {
+    return this._next();
```
amccloud:
```suggestion
    return this._next(first);
```

src/AbstractAPI/index.js
```diff
@@ -279,27 +292,30 @@ export default class AbstractAPI implements AbstractInterface {
...
+            commitSha: objectDescriptor.sha,
+            fileId: objectDescriptor.fileId,
+            layerId: objectDescriptor.layerId,
+            pageId: objectDescriptor.pageId,
+            projectId: objectDescriptor.projectId
+          });
+          const response = await this.fetch(`comments?${query}`);
+          return await response.json();
+        }
+      ).promise;
```
amccloud: 🙌 such a nice diff

src/AbstractAPI/Cursor.js
amccloud: do you think we could make the constructor return this type?

src/AbstractAPI/Cursor.js
tommoor: My read is that the only reason this method exists is to ensure that the first parameter is never passed when called externally?

bitpshr: Ah, yes. This method shouldn't expose any parameter. The type annotation on the public `next` was leftover from before I refactored. Fixed!

src/AbstractAPI/Cursor.js
bitpshr: Great idea. I didn't think this was possible in Flow, but it totally is.

src/AbstractAPI/index.js
bitpshr: Your suggestion to cast the constructor's return type made it even nicer 💪

src/AbstractAPI/index.js
```diff
@@ -279,27 +292,30 @@ export default class AbstractAPI implements AbstractInterface {
...
-        commitSha: objectDescriptor.sha,
-        fileId: objectDescriptor.fileId,
-        layerId: objectDescriptor.layerId,
-        pageId: objectDescriptor.pageId,
-        projectId: objectDescriptor.projectId
-      });
-      const response = await this.fetch(`comments?${query}`);
-      const comments = await unwrapEnvelope(response.json());
-      return comments;
+      return new Cursor<Comment[]>(
```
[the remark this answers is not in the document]
bitpshr: I'm not sure I follow; wouldn't an explicit typing of the request function (and then extracting this and passing it in as the generic) be the same as passing the type in as the generic directly? Candidly, I'm not very familiar with `$Call`, so my understanding may be off.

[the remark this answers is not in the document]
amccloud: Yea, sorry disregard my generic comment. I think you would still need the generic, but I do believe you can move the responsibility of the type to the handler and not the caller by using $Call to define T

something like:

`T: $Call<<R>(() => Promise<R>) => R, request>`

should extracts `R` from the request function

bitpshr:
Thanks for the review so far @amccloud, this should be ready for another pass.

src/AbstractAPI/index.js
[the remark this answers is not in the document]
bitpshr: The only approach I could get to work was to dynamically define a type internally inside `Cursor`, but I wasn't able to then use this type as fully as a generic, so some type safety was ultimately lost.

src/AbstractAPI/Cursor.js
```diff
@@ -0,0 +1,46 @@
+/* @flow */
+import type { CursorMeta, CursorPromise, CursorResponse } from "../";
+
+export default class Cursor<T> {
+  lastResponse: ?Promise<CursorResponse<T> | void>;
+  meta: CursorMeta;
+  request: (meta?: CursorMeta) => Promise<CursorResponse<T>>;
+
+  _next(first?: boolean): CursorPromise<T> {
```
amccloud: nit: thoughts on moving this below constructor? maybe next to `next`?

bitpshr: Good call. I moved this method below the constructor but above `then` and `next` since both methods rely on it.

PR #50, opened 2018-12-17
Allow accessToken option to be a function

rickharris:
Current implementation only accepts a string:

```js
import * as Abstract from "abstract-sdk";

const abstract = Abstract.Client({
  accessToken: process.env.ABSTRACT_API_TOKEN
});
```

New implementation accepts a function that returns a string:

```js
import * as Abstract from "abstract-sdk";

const abstract = Abstract.Client({
  accessToken: () => Cookies.get("authToken")
});
```

Or a function that returns a promise of a string:

```js
import * as Abstract from "abstract-sdk";

const abstract = Abstract.Client({
  accessToken: () => keytar.findPassword("abstract")
});
```

This is kind of a breaking change because the public-facing `Client` object has a public (but undocumented?) `accessToken` property that is currently a string but as of this PR is a function.

Should I add these use cases to the documentation?

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -25,7 +25,7 @@ const abstract = Abstract.Client({
 });
 ```
 
-You pass a function that returns a token.
+You can pass a function that returns a token.
 
 ```js
 import * as Abstract from "abstract-sdk";
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -25,6 +25,28 @@ const abstract = Abstract.Client({
 });
 ```
 
+You pass a function that returns a token.
+
+```js
+import * as Abstract from "abstract-sdk";
+
+const abstract = Abstract.Client({
+  accessToken: () => Cookies.get("abstractToken")
+});
+```
+
+And you can pass a function that returns a promise to a token.
+
+```js
+import * as Abstract from "abstract-sdk";
+
+const abstract = Abstract.Client({
+  accessToken: () => {
+    return new Promise(resolve => resolve(Cookies.get("abstractToken")));
+  }
+});
+```
+
 ## Transports
 
 If you want to ensure that the SDK only ever loads data from the API or the CLI then you can achieve this by specifying a transport. This is useful when running in an environment without the Mac application installed or alternatively when you want to ensure you're only dealing with local data:
--- a/src/AbstractAPI/index.js
+++ b/src/AbstractAPI/index.js
@@ -84,14 +84,19 @@ export default class AbstractAPI implements AbstractInterface {
       ? this._optionAccessToken
       : await this._optionAccessToken();
 
+  async tokenHeader() {
+    const accessToken = await this.accessToken();
+    return accessToken ? { Authorization: `Bearer ${accessToken}` } : {};
+  }
+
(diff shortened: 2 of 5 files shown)


src/AbstractAPI/index.js
```diff
@@ -73,17 +73,22 @@ export default class AbstractAPI implements AbstractInterface {
       );
     }
 
-    this.accessToken = accessToken;
+    this._optionAccessToken = accessToken;
```
rickharris: The intention is that this is a private value, and is prefixed with `option` so that it's clear that this is the thing that was passed in.

src/AbstractCLI/index.js
```diff
@@ -70,12 +70,16 @@ export default class AbstractCLI implements AbstractInterface {
...
+    typeof this._optionAccessToken === "string"
+      ? this._optionAccessToken
+      : await this._optionAccessToken();
+
   async spawn(args: string[]) {
-    return new Promise((resolve, reject) => {
-      const userToken = this.accessToken
-        ? ["--user-token", this.accessToken]
-        : [];
+    const accessToken = await this.accessToken();
```
rickharris: This was moved up in scope because the function inside `new Promise()` is not async. Probably a lot of stuff in that function can be moved to scope, but for the purpose of this PR I only moved what was necessary.

src/index.js.flow
```diff
@@ -1248,11 +1248,14 @@ export type Notification =
   | NotificationUpdateCommit;
 
 export interface AbstractInterface {
-  accessToken: string;
+  accessToken: () => Promise<string>;
```
rickharris: This is the only real change in this diff, everything else was autofixed.

src/index.js.flow
```diff
@@ -24,7 +24,7 @@ export type ActivityDescriptor = {|
 
 export type NotificationDescriptor = {|
   notificationId: string
-|}
+|};
```
rickharris: I have no explanation for why things were out of date, maybe it's because I opened this in my editor and saved it instead of running `yarn lint --fix`, which maybe skips this file since it's not a `.js` file?

tommoor:
Could you update the documentation to reflect this new option also?

src/AbstractAPI/index.js
```diff
@@ -33,7 +33,7 @@ const logStatusSuccess = log.extend("AbstractAPI:status:success");
 const logFetch = log.extend("AbstractAPI:fetch");
 
 export type Options = {
-  accessToken: string,
+  accessToken: string | (() => string | Promise<string>),
```
bitpshr: It may be useful to define a type for `string | (() => string | Promise<string>)` since it's used in a few places.

rickharris: Done in 
https://github.com/goabstract/abstract-sdk/pull/50/commits/7161ad35071fdf026cc9efde449f65f25a558eac

docs/configuration.md
```diff
@@ -25,6 +25,28 @@ const abstract = Abstract.Client({
 });
 ```
 
+You pass a function that returns a token.
```
bitpshr:
```suggestion
You can pass a function that returns a token.
```

PR #51, opened 2018-12-18
Copy flow files so that the module returns typed values

rickharris:
We've had a bit of conversation about how we should enable external flow type checking when using this module, and although we're not in agreement quite yet, I am opening this PR for discussion of its merits and because it unblocks me in trying to incorporate this module into our projects.

### The problem

By using `index.js.flow` as we would normally use a `types.js` file for all of our types, we end up _exporting_ our types, but they're not actually used when we import non-types. For instance, if you import

```js
import * as Abstract from "abstract-sdk";
```

`Abstract` is `any`.

### The possible solution in this PR

Move the current `index.js.flow` to `types.js` and use [`flow-copy-source`](https://www.npmjs.com/package/flow-copy-source) to copy uncompiled source code to `lib`, suffixed with `.flow`. This way, when you import from `abstract-sdk`, the actual code comes from `.js` files, but flow knows to look in their corresponding `.js.flow` files for typings. This is a really simple way to tell flow the types of the things that come from your module.

Now we have type checking and even completions working

image
image

### Tracking some issues using the module with flow enabled

- [ ] By returning `AbstractInterface` from `Abstract.Client`, you can't actually call something like `client.projects.list`, because `projects` is a maybe on `AbstractInterface`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,8 @@
   },
   "scripts": {
     "prepublish": "yarn run build",
-    "build": "babel ./src -d ./lib && flow-copy-source src lib",
+    "build": "babel ./src --ignore **/*/__mocks__,**/test.js -d ./lib",
+    "prebuild": "flow-copy-source --ignore '**/*/__mocks__/*.js' --ignore '**/*test.js' ./src ./lib",
     "lint": "eslint .",
     "test": "jest",
     "test-inspect": "node --inspect-brk ./node_modules/.bin/jest --runInBand",
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
   "main": "lib/index.js",
   "scripts": {
     "prepublish": "yarn run build",
-    "build": "babel ./src -d ./lib",
+    "build": "babel ./src -d ./lib && flow-copy-source src lib",
     "lint": "eslint .",
     "test": "jest",
     "test-inspect": "node --inspect-brk ./node_modules/.bin/jest --runInBand",
@@ -46,6 +46,7 @@
     "eslint-plugin-prettier": "^2.6.2",
     "eslint-plugin-react": "^7.11.1",
     "flow-bin": "^0.80.0",
+    "flow-copy-source": "^2.0.2",
     "flow-typed": "^2.5.1",
     "jest": "^23.6.0",
     "jest-fetch-mock": "GrahamTheCoder/jest-fetch-mock#blob-fix",
--- a/src/types.js
+++ b/src/types.js
@@ -38,12 +38,12 @@ export type CommitDescriptor = {|
 
 export type BranchDescriptor = {|
   ...ObjectDescriptor,
-  sha: $PropertyType<ObjectDescriptor, "sha"> | "latest"
+  sha: $PropertyType<ObjectDescriptor, "sha">
 |};
 
 export type FileDescriptor = {|
   ...ObjectDescriptor,
-  sha: $PropertyType<ObjectDescriptor, "sha"> | "latest",
+  sha: $PropertyType<ObjectDescriptor, "sha">,
   fileId: string
 |};
 
@@ -55,7 +55,7 @@ export type PageDescriptor = {|
 
 export type LayerDescriptor = {|
   ...ObjectDescriptor,
-  sha: $PropertyType<ObjectDescriptor, "sha"> | "latest",
+  sha: $PropertyType<ObjectDescriptor, "sha">,
   fileId: string,
   pageId: string,
   layerId: string
--- a/src/types.js
+++ b/src/types.js
@@ -24,7 +24,7 @@ export type ActivityDescriptor = {|
 
 export type NotificationDescriptor = {|
   notificationId: string
-|}
+|};
 
 type ObjectDescriptor = {|
   projectId: string,
@@ -1252,7 +1252,10 @@ export interface AbstractInterface {
 
   activities?: {
     list: (
-      objectDescriptor?: BranchDescriptor | OrganizationDescriptor | ProjectDescriptor,
+      objectDescriptor?:
+        | BranchDescriptor
+        | OrganizationDescriptor
+        | ProjectDescriptor,
       options?: ListOptions
     ) => Promise<Activity[]>,
     info: (activityDescriptor: ActivityDescriptor) => Promise<Activity>
@@ -1290,7 +1293,9 @@ export interface AbstractInterface {
     list: (
       {
         projectId: $PropertyType<ProjectDescriptor, "projectId">
-      } & $Shape<BranchDescriptor & CommitDescriptor & PageDescriptor & LayerDescriptor>,
+      } & $Shape<
+        BranchDescriptor & CommitDescriptor & PageDescriptor & LayerDescriptor
+      >,
       options?: ListOptions
     ) => Promise<Comment[]>,
     info: (commentDescriptor: CommentDescriptor) => Promise<Comment>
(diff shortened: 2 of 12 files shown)


bitpshr:
Thanks for putting this together, it's good to see how this approach looks in practice. It's very surprising to me that the Flow compiler doesn't have any option to generate public types for each module it encounters. In TypeScript, `.d.ts` files are placed alongside compiled `.js` files by default, making both type generation and consumption mostly automatic. I noticed early versions of Flow included the [seemingly-broken](https://github.com/facebook/flow/issues/5871) `gen-flow-files` command, but it doesn't appear to be explicitly supported or documented anymore.

One issue with the `flow-copy-source` approach is resulting module size. By copying uncompiled source code instead of generating minimal type definitions, we're increasing the resulting package size in a non-negligible manner to appease downstream compilers.

Thinking out loud: how do other major production projects generate typings for their compiled code using Flow? This seems like an incredibly important feature for Flow to omit. Imagine a massive project with hundreds of files that will ultimately be distributed via NPM; are the two options really to copy entire source trees or to hand-roll types?

rickharris:
We might take a look at how React does it, as they're probably the biggest project that does Flow typing. Although since they're Facebook, too, they might even have `React` types built into flow at this point :\

bitpshr:
What about [using Babel](https://babeljs.io/docs/en/babel-plugin-transform-flow-comments) to convert type annotations to comments? This may be a happy medium: we only ship built code, but the built code contains commented-out type annotations that are still picked up by Flow. I'd personally be OK with uglier (and larger) compiled code if it included type annotations, especially since the type comments could be stripped out in downstream projects as long as they're built (regardless of whether they use Flow.)

rickharris:
So I gave `@babel/plugin-transform-flow-comments` a shot and... it doesn't work.

First thing I noticed - the compiled files still don't have a `@flow` pragma, so Flow doesn't even know to type check them. I have no idea why that plugin wouldn't add the flow pragma, but that's not the worst of it.

Secondly, it seems like flow comments either don't work with _all_ flow code, or either the plugin itself doesn't cover all of flow's language feature. After compiling `abstract-sdk` with this plugin enabled, it spit out code with syntax errors:

```
ERROR in ./node_modules/@elasticprojects/abstract-core/node_modules/abstract-sdk/lib/AbstractAPI/index.js
Module parse failed: Unexpected token (68:18)
You may need an appropriate loader to handle this file type.
| }
|
| class AbstractAPI implements {
|   constructor({
|     /*: Options*/
```

The `implements` keyword is a flow thing and isn't getting handled properly here.

amccloud:
Not sure I see the need to move away from index.js.flow

Could we add a libdef to it and make sure it gets bundled in the package?

Most and Ava do something similar:

- https://github.com/cujojs/most/blob/master/src/index.js.flow
- https://github.com/avajs/ava/blob/master/index.js.flow

rickharris:
I'd like to make one last argument in favor of this PR & `flow-copy-source`: the libdef is going to be harder to right, and this method is _so easy_ and it just works. The only downfall of moving `index.js.flow` to `types.js` and using `flow-copy-source` is that it does double the amount of code in the published npm module, but two things to keep in mind:

- It doesn't change the _compiled size_ of code that uses it
- For now, at the very least, this really isn't a problem because this project is not that big.

cc @amccloud @bitpshr

amccloud:
@rickharris seems worth it if you get a libdef for free! Can we ever opt-out of publishing a type or is any type we export internally also a public type because it will be copied?

rickharris:
> Can we ever opt-out of publishing a type or is any type we export internally also a public type because it will be copied?

That's a good question. We could control what gets exported from the index. So say that we have `src/types.js` and it has a type called `AbstractInternalThing` and we don't really want to publish it, you couldn't import it like `import type { AbstractInternalThing } from 'abstract-sdk'`. But you could import it as `import type { AbstractInternalThing } from 'abstract-sdk/lib/types'`, if someone read the source code and decided they wanted to do that. Usually that kind of thing (things inside of `lib` directories) in modules is considered private/internal. If we specifically _did_ want to export a certain type, we could `export type { AbstractExternalThing } from './types'` and that would do the trick.

It might also be possible to only copy `index.js` using `flow-copy-source` such that only the public-facing API was available for being typed, but I'm not sure if that would 100% work or not. Like say inside of `index.js` we have `export { Client } from './Client'`, then I'm not sure `Client` would be typed unless we also copied `Client/index.js` as `Client/index.js.flow`. We could test that out.

rickharris:
@amccloud do you have thoughts on the issue I described in the main PR description?

> By returning `AbstractInterface` from `Abstract.Client`, you can't actually call something like `client.projects.list`, because `projects` is a maybe on `AbstractInterface`

amccloud:
@rickharris that was sorta intentional because depending on the transport the endpoint may not be available. We could make AbstractCLI implement everything but stub/throw if something not supported by cli is called?

rickharris:
🤔 seems less than ideal for the default AUTO case which will have everything. Assuming that `AbstractInterface` was mostly meant to be the interface that the transports implement, we could probably return the actual type of whatever transport is returned from `Abstract.Client` so that a particular function can be either defined or undefined on the transport, but not a maybe.

rickharris:
Had a convo with @amccloud and there's still a lot to figure out about our exported types for this package, but in the meantime we might as well merge this so we have types for the existing code.

package.json
```diff
@@ -12,7 +12,7 @@
   },
   "scripts": {
     "prepublish": "yarn run build",
-    "build": "babel ./src -d ./lib",
+    "build": "babel ./src -d ./lib && flow-copy-source src lib",
```
amccloud: would be nice to switch the order of these so the babel options like `--watch` can still be used with build

amccloud: also nit: ./src ./lib to match all the other relative paths

rickharris: Fair point. We'd probably want a kind of `watch` command to do both of these things, wouldn't we?

amccloud: yea something like that or like `start` for development

amccloud:
@rickharris descriptors are public API https://sdk.goabstract.com/docs/abstract-api/#descriptors

Individual properties like `projectId`, `branchId`, `sha`, etc. are now internals, so components should no longer depend on those

rickharris:
@amccloud I'm confused about what points you're making here. Descriptors are still just objects of string ids, like you said in your edit, right?

amccloud:
@rickharris descriptors are now valid dependencies too so components should prefer to depend on descriptors if they can:

```js
type Props = {
 params: Abstract.LayerDescriptor,
 layer: Abstract.Layer // would be nice too
};
```

src/types.js
```diff
@@ -55,7 +55,7 @@ export type PageDescriptor = {|
 
 export type LayerDescriptor = {|
   ...ObjectDescriptor,
-  sha: $PropertyType<ObjectDescriptor, "sha"> | "latest",
+  sha: $PropertyType<ObjectDescriptor, "sha">,
```
amccloud: Adding a note here for future us. Flow doesn't appear to be refining the literal "latest" when a desscriptor (like `CommitDescriptor`) and descriptor that supports latest are in a union

screenshot 2019-01-08 15 53 42

amccloud:
@rickharris I really think we should export all of the types that are documented and intended to be public API. It would be the types in the right sidebar here  https://sdk.goabstract.com/docs/abstract-api/ and the subtypes we document like `ChangesetChange`, `CollectionLayer`, `Annotation`, etc.

This PR could benefit from the types https://github.com/goabstract/ui/pull/202 and it was the intention all along with `index.js.flow`

bitpshr:
I agree with @amccloud. From a high level, consumers of a typed API should be able to use the same interfaces that make up that public API.

rickharris:
@amccloud check out cfc83bb and 2760ed1. I think this should be good to go now.

PR #58, opened 2018-12-18
Fix issues with accessToken option function

rickharris:
- [x] `Abstract.Client` didn't accept the new `AccessTokenOption` type even though the underlying transports did.

### DIFF
--- a/src/AbstractAPI/index.js
+++ b/src/AbstractAPI/index.js
@@ -79,13 +79,10 @@ export default class AbstractAPI implements AbstractInterface {
     this.previewsUrl = previewsUrl;
   }
 
-  accessToken = async () => {
-    if (!this._optionAccessToken) return;
-    if (typeof this._optionAccessToken === "string") {
-      return this._optionAccessToken;
-    }
-    return this._optionAccessToken();
-  };
+  accessToken = async () =>
+    typeof this._optionAccessToken === "function"
+      ? this._optionAccessToken()
+      : this._optionAccessToken;
 
   async tokenHeader() {
     const accessToken = await this.accessToken();
--- a/src/AbstractAPI/index.js
+++ b/src/AbstractAPI/index.js
@@ -79,10 +79,13 @@ export default class AbstractAPI implements AbstractInterface {
     this.previewsUrl = previewsUrl;
   }
 
-  accessToken = async () =>
-    typeof this._optionAccessToken === "string"
-      ? this._optionAccessToken
-      : await this._optionAccessToken();
+  accessToken = async () => {
+    if (!this._optionAccessToken) return;
+    if (typeof this._optionAccessToken === "string") {
+      return this._optionAccessToken;
+    }
+    return this._optionAccessToken();
+  };
 
   async tokenHeader() {
     const accessToken = await this.accessToken();
--- a/src/Client/index.js
+++ b/src/Client/index.js
@@ -13,11 +13,15 @@ type Options = {
 export default function Client(options: Options = {}): AbstractInterface {
   const Transport = options.transport || AUTO;
   const accessToken = options.accessToken || process.env.ABSTRACT_TOKEN;
-  const { apiUrl, cliPath, previewsUrl } = options;
 
   if (!Transport) {
     throw new Error("options.transport is required");
   }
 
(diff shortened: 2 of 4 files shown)


src/AbstractAPI/index.js
```diff
@@ -79,10 +79,13 @@ export default class AbstractAPI implements AbstractInterface {
     this.previewsUrl = previewsUrl;
   }
 
-  accessToken = async () =>
-    typeof this._optionAccessToken === "string"
-      ? this._optionAccessToken
-      : await this._optionAccessToken();
+  accessToken = async () => {
+    if (!this._optionAccessToken) return;
```
rickharris: Since `this._optionAccessToken` can be `?string` now.

src/Client/index.js
```diff
@@ -1,22 +1,20 @@
 // @flow
-import type { AbstractInterface } from "../";
+import type { AbstractInterface, AccessTokenOption } from "../";
 import { AUTO } from "../transports";
 
 type Options = {
-  accessToken?: string,
+  accessToken?: AccessTokenOption,
```
rickharris: 🍖

src/Client/index.js
```diff
@@ -1,22 +1,20 @@
...
-export default function Client({
-  accessToken = process.env.ABSTRACT_TOKEN || "",
-  cliPath,
-  apiUrl,
-  previewsUrl,
-  transport: Transport = AUTO
-}: Options = {}): AbstractInterface {
+export default function Client(options: Options = {}): AbstractInterface {
+  const Transport = options.transport || AUTO;
+  const accessToken = options.accessToken || process.env.ABSTRACT_TOKEN;
```
rickharris: There were weird type errors around `accessToken` when de-structuring after making the 🍖 change in this PR. Doing it this way cleared the errors up.

Also no need for `process.env.ABSTRACT_TOKEN || ""` here since we have to embrace `?string` as the return type for `accessToken`.

tommoor: This is honestly easier to read for me too 🙈

src/AbstractAPI/index.js
```diff
@@ -79,10 +79,13 @@ export default class AbstractAPI implements AbstractInterface {
...
+    if (typeof this._optionAccessToken === "string") {
+      return this._optionAccessToken;
+    }
+    return this._optionAccessToken();
```
tommoor: `await` purposefully lost here?

rickharris: Yeah, it's unnecessary since this function itself has to be awaited anyway.

src/Client/index.js
rickharris: Yeah, I think I agree.

src/AbstractAPI/index.js
rickharris: We could, for sure, although that changes the return type. Current it's something like `Promise<undefined | string>`, and removing the `async` keyword would make it more like: `undefined | string | Promise<string>`, and I thought the former would be more straightforward.

tommoor: aha, gtk!

src/Client/index.js
amccloud: Can we take this approach to destructuring everywhere 😆

Was used here because of the defaults or in other cases because of re-assignment. Otherwise, I tend to opt out of using destructuring altogether

src/AbstractAPI/index.js
amccloud: Can we check if `this._optionAccessToken` is callable, call it, otherwise return it?

rickharris: That sounds nice, how do you check for call-ability?

rickharris: Updated in 
https://github.com/goabstract/abstract-sdk/pull/58/commits/3b886945f57c6493f685ad15ad28d3be623238f3

src/Client/index.js
```diff
@@ -1,22 +1,20 @@
...
+  const { apiUrl, cliPath, previewsUrl } = options;
```
amccloud: @rickharris nit: would prefer to avoid creating variables just for unchanged values. would be nice to continue to use `options.apiUrl` and etc below

Issue #55, opened 2018-12-18
Add users endpoints to API transport


tommoor:
We're currently missing endpoints for `users.list` and `users.info` – as we don't send this information back with branch data (https://github.com/goabstract/abstract-sdk/issues/45) it's important that it can be loaded separately.

bitpshr:
Would we use the `/scim/v2/Users` endpoints for this, or would we add new endpoints to the users controller exposed at `/users`?

tommoor:
Nothing related to SCIM here – thanks for asking for clarification. This is an interesting one now that I remember our API in more detail, right now we have endpoints for "memberships", but nothing for users, eg:

`GET /organizations/:organization_id/memberships`
`GET /projects/:project_id/memberships`

These return membership objects that include a nested `user`:

![image](https://user-images.githubusercontent.com/380914/50188763-8be86c00-02d7-11e9-8a70-275c957c56fd.png)

I'm not entirely sure how we should represent this in the SDK now and very open to thoughts…

I do think it would be worth adding a `GET /users/:id` too for individual user lookup though.

bitpshr:
What if `users.list` took either an `OrganizationDescriptor` or a `ProjectDescriptor`, and depending on which one is passed, it calls out to either `/organizations/:organization_id/memberships` or `/projects/:project_id/memberships`, respectively. It could map the resulting membership list and only return users.

Alternatively, I'd be fine with adding a new `GET /users` endpoint too. I think we'd just find the organization based on an `organization_id` URL parameter, then return its `users`.

PR #61, opened 2018-12-19
feature: add users.list and users.info

bitpshr:
This pull request adds support for `users.list` and `users.info` to the API transport.

Requires https://github.com/goabstract/projects/pull/1133
Resolves https://github.com/goabstract/abstract-sdk/issues/55

### DIFF
--- a/docs/abstract-api.md
+++ b/docs/abstract-api.md
@@ -868,7 +868,7 @@ const branchFiles = await abstract.files.list(branchShare.descriptor);
 
 ## Users
 
-A user contains information specific to an individual account. Users are created by creating a new account.
+A user contains information specific to an individual account. Users are global to Abstract and are not specific to organizations. A user is created in the application by creating a new account.
 
 ![API][api-icon]
 
@@ -883,7 +883,6 @@ A user contains information specific to an individual account. Users are created
 | `id`              | `string` | UUID identifier for the user                                 |
 | `name`            | `string` | The name of the page                                         |
 | `primaryEmailId`  | `string` | ID of the primary email for this user                        |
-| `releaseChannel`  | `string` | Release channel this user belongs to                         |
 | `updatedAt`       | `string` | Timestamp indicating when this account was updated           |
 | `username`        | `string` | Username associated with this user                           |
 
@@ -1021,7 +1020,5 @@ Reference for the parameters required to load resources with Abstract SDK.
 ### UserDescriptor
 
  ```js
-{
-  userId: string
-}
+{ userId: string }
 ```
--- a/docs/abstract-api.md
+++ b/docs/abstract-api.md
@@ -866,6 +866,52 @@ const branchFiles = await abstract.files.list(branchShare.descriptor);
 ```
 
 
+## Users
+
+A user contains information specific to an individual account. Users are created by creating a new account.
+
+![API][api-icon]
+
+### The user object
+
+| Property          | Type     | Description                                                  |
+|-------------------|----------|--------------------------------------------------------------|
+| `avatarUrl`       | `string` | URL of the avatar for this user                              |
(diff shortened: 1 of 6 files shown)


bitpshr:
Note: The upstream PR at https://github.com/goabstract/projects/pull/1133 landed so this should be ready for review.

docs/abstract-api.md
```diff
@@ -866,6 +866,52 @@ const branchFiles = await abstract.files.list(branchShare.descriptor);
...
+| Property          | Type     | Description                                                  |
+|-------------------|----------|--------------------------------------------------------------|
+| `avatarUrl`       | `string` | URL of the avatar for this user                              |
+| `createdAt`       |  `string`| Timestamp indicating when this account was created           |
+| `deletedAt`       | `string` | Timestamp indicating when this account was deleted           |
+| `email`           | `string` | Email associated with this user account                      |
+| `id`              | `string` | UUID identifier for the user                                 |
+| `name`            | `string` | The name of the page                                         |
+| `primaryEmailId`  | `string` | ID of the primary email for this user                        |
+| `releaseChannel`  | `string` | Release channel this user belongs to                         |
```
tommoor: I'm tempted to say we should remove this, and remove from the API responses too – seems a little leaky of internal flags.

docs/abstract-api.md
```diff
@@ -866,6 +866,52 @@ const branchFiles = await abstract.files.list(branchShare.descriptor);
...
+
+| Property          | Type     | Description                                                  |
+|-------------------|----------|--------------------------------------------------------------|
+| `avatarUrl`       | `string` | URL of the avatar for this user                              |
+| `createdAt`       |  `string`| Timestamp indicating when this account was created           |
+| `deletedAt`       | `string` | Timestamp indicating when this account was deleted           |
+| `email`           | `string` | Email associated with this user account                      |
+| `id`              | `string` | UUID identifier for the user                                 |
+| `name`            | `string` | The name of the page                                         |
+| `primaryEmailId`  | `string` | ID of the primary email for this user                        |
```
tommoor: This one isn't going to make much sense without exposing the `emails` endpoints too, but probably fine for now 😄

docs/abstract-api.md
bitpshr: Actually, after digging through the backend and frontend monorepos, nowhere actually uses `releaseChannel` in the returned JSON. Would it be preferable to remove this property from the JSON serializer altogether so it never gets returned when a user is returned? This may be too restrictive.

docs/abstract-api.md
```diff
@@ -971,3 +1017,11 @@ Reference for the parameters required to load resources with Abstract SDK.
   commentId: string
 }
 ```
+
+### UserDescriptor
+
+ ```js
+{
+  userId: string
+}
```
amccloud: nit: one line `{ userId: string }`?

src/AbstractAPI/index.js
```diff
@@ -567,6 +568,26 @@ export default class AbstractAPI implements AbstractInterface {
     }
   };
 
+  users = {
+    list: async (
+      objectDescriptor: OrganizationDescriptor | ProjectDescriptor
+    ) => {
+      let url = "";
```
amccloud: const and ternary over reassignment?

docs/abstract-api.md
tommoor: You know I thought we were using it still in desktop and we'd have to make some sort of exception – but it all seems to be based on the feature flags now.

Lets remove from the serializer too!

bitpshr: Removed from the SDK documentation. Remote tracking for upstream changes:

https://github.com/goabstract/projects/pull/1148
https://github.com/goabstract/ui/pull/157

Issue #54, opened 2018-12-18
Add branches endpoints to CLI transport


tommoor:
Loading local branches is a definite requirement.

https://sdk.goabstract.com/docs/reference/#branches

PR #64, opened 2019-01-05
feature: add branches.info and branches.list to CLI

bitpshr:
This pull request adds `branches.info` and `branches.list` support to the CLI transport.

Requires https://github.com/goabstract/projects/pull/1199
Resolves https://github.com/goabstract/abstract-sdk/issues/54

### DIFF
--- a/src/AbstractCLI/index.js
+++ b/src/AbstractCLI/index.js
@@ -317,4 +317,21 @@ export default class AbstractCLI implements AbstractInterface {
       ]);
     }
   };
+
+  branches = {
+    list: async (projectDescriptor: ProjectDescriptor) => {
+      const data = await this.spawn(["branches", projectDescriptor.projectId]);
+      return data.branches;
+    },
+    info: async (branchDescriptor: BranchDescriptor) => {
+      const data = await this.spawn([
+        "branch",
+        "load",
+        branchDescriptor.branchId,
+        branchDescriptor.projectId
+      ]);
+
+      return data;
+    }
+  };
 }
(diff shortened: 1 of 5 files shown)


tommoor:
Sweet! I think we should get the `active` filter in here before merging and releasing though

bitpshr:
Thanks @tommoor. Upstream PR at goabstract/projects#1199 updated to support `filter`.

src/AbstractCLI/index.js
```diff
@@ -317,4 +317,32 @@ export default class AbstractCLI implements AbstractInterface {
...
+
+      const data = await this.spawn([
+        "branches",
+        projectDescriptor.projectId,
+        ...branchesArgs
+      ]);
+      return data.branches;
+    },
+    info: async (branchDescriptor: BranchDescriptor) => {
+      const data = await this.spawn([
```
tommoor: I think we can just `return this.spawn(...` in these cases where no further processing needs to happen

src/AbstractCLI/index.js
```diff
@@ -317,4 +317,32 @@ export default class AbstractCLI implements AbstractInterface {
...
     }
   };
+
+  branches = {
+    list: async (
+      projectDescriptor: ProjectDescriptor,
+      options: { filter?: "active" | "archived" | "mine" } = {}
+    ) => {
+      const branchesArgs = options.filter
+        ? ["--branch-filter", options.filter]
```
tommoor: Belongs on the other PR I'm sure, but wondering why the filter also has "branch" when the command is already "branches", `--filter` would be more appropriate?

bitpshr: Upstream PR updated.

bitpshr: This PR has also been updated now, so everything should be using the new `filter` flag. Ready for another review when you have a chance.

PR #65, opened 2019-01-07
feature: add assets endpoints to the API transport

### LINKED ISSUE #52: Add assets endpoints
- [ ] `assets.list`
- [ ] `assets.info`
- [ ] `assets.raw`

bitpshr:
This pull request updates the API transport by adding `assets`-related endpoints.

Requires https://github.com/goabstract/projects/pull/1200
Resolves #52

### DIFF
--- a/docs/abstract-api.md
+++ b/docs/abstract-api.md
@@ -6,6 +6,7 @@ title: Client
 [cli-icon]: https://img.shields.io/badge/CLI-lightgrey.svg
 [api-icon]: https://img.shields.io/badge/API-blue.svg
 
+
 ## Activities
 
 ![API][api-icon]
@@ -47,6 +48,82 @@ abstract.activities.info({
   activityId: "616daa90-1736-11e8-b8b0-8d1fec7aef78"
 });
 ```
+
+
+[cli-icon]: https://img.shields.io/badge/CLI-lightgrey.svg
+[api-icon]: https://img.shields.io/badge/API-blue.svg
+
+
+## Assets
+
+![API][api-icon]
+
+An asset represents an exportable resource from a Sketch file. Files are automatically scanned for new assets as new commits are pushed to a project.
+
+### The asset object
+
+| Property      | Type                | Description                                                                             |
+|---------------|---------------------|-----------------------------------------------------------------------------------------|
+| `createdAt`   | `string`            | Timestamp at which the asset was created                                                |
+| `defaultAbstractFormat` | `boolean` | Indicates if this asset is the default asset for a layer                                |
+| `fileFormat`  | `string`            | File type of this asset                                                                 |
+| `fileId`      | `string`            | File ID of this asset                                                                   |
+| `formatName`  | `string`            | Format of this file, e.g. "2x"                                                          |
+| `id`          | `string`            | UUID identifier of the asset                                                            |
+| `layerId`     | `string`            | UUID of the layer this asset belongs to                                                 |
+| `layerName`   | `string`            | Name of the layer this asset belongs to                                                 |
+| `namingScheme`| `string`            | Indicates the naming convention used for this asset                                     |
+| `nestedLayerId` | `string`          | ID of the nested layer this asset belongs to                                            |
+| `projectId`   | `string`            | ID of the project this layer belongs to                                                 |
+| `scale`       | `string`            | Scale of this asset in pixels                                                           |
+| `sha`         | `string`            | SHA of the commit containing the version of the file this asset belongs to              |
+| `url`         | `string`            | Direct URL to the asset file                                                            |
+
+### List all assets
+
+`assets.list(BranchDescriptor): Promise<Asset[]>`
+
+List all assets for a commit on a given branch of a project
+
+```js
+abstract.assets.list({
+  branchId: "8a13eb62-a42f-435f-b3a3-39af939ad31b",
+  projectId: "b8bf5540-6e1e-11e6-8526-2d315b6ef48f",
+  sha: "c4e5578c590f5334349b6d7f0dfd4d3882361f1a", // or sha: "latest"
+});
+```
+
+### Retrieve an asset
+
+`asset.info(AssetDescriptor): Promise<Asset>`
+
+Load the info for an asset
+
+```js
+abstract.assets.info({
+  assetId: "fcd67bab-e5c3-4679-b879-daa5d5746cc2",
+  projectId: "b8bf5540-6e1e-11e6-8526-2d315b6ef48f"
+});
+```
+
+### Retrieve an asset file
+
+`asset.raw(AssetDescriptor): Promise<ArrayBuffer>`
+
+Load given asset file based on its ID. The resulting `ArrayBuffer` can be used with node `fs` APIs. For example, it's possible to write the image to disk:
+
+```js
+const arrayBuffer = abstract.assets.raw({
+  assetId: "fcd67bab-e5c3-4679-b879-daa5d5746cc2",
+  projectId: "b8bf5540-6e1e-11e6-8526-2d315b6ef48f"
+});
+
+fs.writeFile("asset.png", Buffer.from(arrayBuffer), (err) => {
+  if (err) throw err;
+  console.log("Asset image written!");
+});
+```
+
 
 ## Branches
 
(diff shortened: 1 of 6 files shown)


docs/abstract-api.md
```diff
@@ -48,6 +49,82 @@ abstract.activities.info({
...
+
+[cli-icon]: https://img.shields.io/badge/CLI-lightgrey.svg
+[api-icon]: https://img.shields.io/badge/API-blue.svg
+
+
+## Assets
+
+![API][api-icon]
+
+An asset represents an exportable resource from a Sketch file. Files are automatically scanned for new assets as new commits are pushed to a project.
```
tommoor: Suggested copy: 

> An asset represents a resource exported from a design file. Assets are automatically updated and available for new commits.

docs/abstract-api.md
```diff
@@ -48,6 +49,82 @@ abstract.activities.info({
...
+| `fileId`      | `string`            | File ID of this asset                                                                   |
+| `formatName`  | `string`            | Format of this file, e.g. "2x"                                                          |
+| `id`          | `string`            | UUID identifier of the asset                                                            |
+| `layerId`     | `string`            | UUID of the layer this asset belongs to                                                 |
+| `layerName`   | `string`            | Name of the layer this asset belongs to                                                 |
+| `namingScheme`| `string`            | Indicates the naming convention used for this asset                                     |
+| `nestedLayerId` | `string`          | ID of the nested layer this asset belongs to                                            |
+| `projectId`   | `string`            | ID of the project this layer belongs to                                                 |
+| `scale`       | `string`            | Scale of this asset in pixels                                                           |
```
tommoor: Could you provide an example of the value here? I'm not sure what I would expect given this description.

docs/abstract-api.md
```diff
@@ -48,6 +49,82 @@ abstract.activities.info({
...
+| `createdAt`   | `string`            | Timestamp at which the asset was created                                                |
+| `defaultAbstractFormat` | `boolean` | Indicates if this asset is the default asset for a layer                                |
+| `fileFormat`  | `string`            | File type of this asset                                                                 |
+| `fileId`      | `string`            | File ID of this asset                                                                   |
+| `formatName`  | `string`            | Format of this file, e.g. "2x"                                                          |
+| `id`          | `string`            | UUID identifier of the asset                                                            |
+| `layerId`     | `string`            | UUID of the layer this asset belongs to                                                 |
+| `layerName`   | `string`            | Name of the layer this asset belongs to                                                 |
+| `namingScheme`| `string`            | Indicates the naming convention used for this asset                                     |
```
tommoor: Similar here, is there an example or two we can give?

docs/abstract-api.md
bitpshr: Examples added for this and `namingScheme`. These properties still aren't super useful and feel a little internal, but they're used downstream in the app UIs.

bitpshr:
Thanks @tommoor. Ready for another pass when you have time.

src/AbstractAPI/index.js
```diff
@@ -611,6 +613,41 @@ export default class AbstractAPI implements AbstractInterface {
...
+      const response = await this.fetch(
+        asset.url,
+        {
+          headers: {
+            Accept: undefined,
+            "Content-Type": undefined,
+            "Abstract-Api-Version": undefined
+          }
+        },
+        ""
```
natatat: is this last empty string argument necessary?

bitpshr: Sort of: this last argument can be a custom `hostname` that overrides the internally-applied default. Passing in an empty string in this case allows us to use this `fetch` method to request an arbitrary URL without any hostname applied at all (e.g. a URL that already includes its hostname, like an asset URL).

Thinking about it more, we can make this a bit cleaner by changing this to be a boolean when no hostname should be used. Updated.

src/AbstractAPI/index.js
```diff
@@ -94,7 +94,11 @@ export default class AbstractAPI implements AbstractInterface {
     return accessToken ? { Authorization: `Bearer ${accessToken}` } : {};
   }
 
-  async fetch(input: string | URL, init: Object = {}, hostname?: string) {
+  async fetch(
+    input: string | URL,
+    init: Object = {},
+    hostname?: string | boolean
```
amccloud: `true` would be an odd option. what about `hostname?: ?string`?

bitpshr: This would still restrict us to either passing a `hostname` or not passing one and using the default; we still need a way to communicate that the hostname should be ignored altogether. I think what we actually want is `hostname?: string | false`.

amccloud: @bitpshr `this.fetch(_, _, null)`?

bitpshr: Explicitly passing `null` instead explicitly passing `false` feels like the same thing to me, but I can switch to that if you think it's a clearer API. Either way, we're passing an extra overloaded parameter to indicate special behavior.

bitpshr: Updated.

bitpshr:
Thanks for the review so far @amccloud and @amccloud. This should be ready for another pass when you have a chance.

docs/abstract-api.md
```diff
@@ -48,6 +49,82 @@ abstract.activities.info({
...
+| `fileId`      | `string`            | File ID of this asset                                                                   |
+| `formatName`  | `string`            | Format of this file, e.g. "2x"                                                          |
+| `id`          | `string`            | UUID identifier of the asset                                                            |
+| `layerId`     | `string`            | UUID of the layer this asset belongs to                                                 |
+| `layerName`   | `string`            | Name of the layer this asset belongs to                                                 |
+| `namingScheme`| `string`            | String indicating the naming convention used for this asset, e.g. "1"                   |
+| `nestedLayerId` | `string`          | ID of the nested layer this asset belongs to                                            |
+| `projectId`   | `string`            | ID of the project this layer belongs to                                                 |
+| `scale`       | `string`            | Scale of this asset in pixels, e.g. "112px"                                             |
```
anolson: `scale` maps to the raw scale property from Sketch, it will look something like `1.00`, `2.00`, etc.

docs/abstract-api.md
```diff
@@ -48,6 +49,82 @@ abstract.activities.info({
...
+| `fileFormat`  | `string`            | File type of this asset                                                                 |
+| `fileId`      | `string`            | File ID of this asset                                                                   |
+| `formatName`  | `string`            | Format of this file, e.g. "2x"                                                          |
+| `id`          | `string`            | UUID identifier of the asset                                                            |
+| `layerId`     | `string`            | UUID of the layer this asset belongs to                                                 |
+| `layerName`   | `string`            | Name of the layer this asset belongs to                                                 |
+| `namingScheme`| `string`            | String indicating the naming convention used for this asset, e.g. "1"                   |
+| `nestedLayerId` | `string`          | ID of the nested layer this asset belongs to                                            |
+| `projectId`   | `string`            | ID of the project this layer belongs to                                                 |
```
anolson: layer -> asset?

docs/abstract-api.md
```diff
@@ -48,6 +49,82 @@ abstract.activities.info({
...
+| `createdAt`   | `string`            | Timestamp at which the asset was created                                                |
+| `defaultAbstractFormat` | `boolean` | Indicates if this asset is the default asset for a layer                                |
+| `fileFormat`  | `string`            | File type of this asset                                                                 |
+| `fileId`      | `string`            | File ID of this asset                                                                   |
+| `formatName`  | `string`            | Format of this file, e.g. "2x"                                                          |
+| `id`          | `string`            | UUID identifier of the asset                                                            |
+| `layerId`     | `string`            | UUID of the layer this asset belongs to                                                 |
+| `layerName`   | `string`            | Name of the layer this asset belongs to                                                 |
+| `namingScheme`| `string`            | String indicating the naming convention used for this asset, e.g. "1"                   |
```
anolson: The `namingScheme` attribute determines how the `formatName` is applied when constructing the filename for an asset – `1 = Prefix, 0 = Suffix`

**Prefix**
```
icons/Star.png
```

**Suffix**
```
Star@2x.png
```

We discussed constructing the filename on the server, but landed on initially doing that at the client.

bitpshr:
Thanks a ton for the added context @anolson. Documentation updated.

src/AbstractAPI/index.js
```diff
@@ -156,6 +160,22 @@ export default class AbstractAPI implements AbstractInterface {
...
+      {
+        ...init,
+        headers: {
+          Accept: undefined,
+          "Content-Type": undefined,
+          "Abstract-Api-Version": undefined,
+          ...init.headers
+        }
+      },
+      ""
```
amccloud: should this be `null` now?

Would you be down with making fetchPreview use fetchRaw?

`return this.fetchRaw(input, init, this.previewsUrl)`

amccloud: Edit: is this still used?

src/AbstractAPI/test.js
```diff
@@ -99,12 +103,7 @@ const responses = {
     ]
   },
   previews: {
-    arrayBuffer: (
-      // inlined to avoid multiple reads
-      data = fs.readFileSync(
-        path.resolve(__dirname, "../../fixtures/preview.png")
-      )
-    ) => [data, { status: 200 }]
+    arrayBuffer: () => [buffer, { status: 200 }]
```
amccloud: This usually matches the name of the action that we're mocking. Can you rename to `raw`?

src/AbstractAPI/index.js
bitpshr: Good catch. Removed.

PR #66, opened 2019-01-07
Browser support

amccloud:
Webpack and other bundlers will respect the ["browser" field in package.json](https://github.com/defunctzombie/package-browser-field-spec) when bundling for web. We can use it to disable AbstractCLI in browsers.

Alternatively, we could supply a mock `AbstractCLIWeb` but removing the module altogether and letting AUTO gracefully degrade seems to work too

### DIFF
--- a/src/AUTO/index.js
+++ b/src/AUTO/index.js
@@ -8,8 +8,10 @@ type Options = OptionsCLI & OptionsAPI;
 // $FlowFixMe
 class AbstractAUTO implements AbstractInterface {
   constructor(options: Options) {
+    // AbstractCLI is not available in browsers (see: package.json#browsers)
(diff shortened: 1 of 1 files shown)


src/AUTO/index.js
```diff
@@ -8,7 +8,8 @@ type Options = OptionsCLI & OptionsAPI;
 // $FlowFixMe
 class AbstractAUTO implements AbstractInterface {
   constructor(options: Options) {
-    const cli = new AbstractCLI(options);
+    const cli =
```
tommoor: I like this approach, I think it worthy of a comment as to why this can be false though as future-us might not make the connection with the webpack config read from package.json 😄

PR #68, opened 2019-01-08
Fix for commits.info(layer|file) at latest

amccloud:
Noticed this as I started to build something with the sdk. This is a holdover from when sha could be `undefined` to represent "latest". No test for this so it was missed 🙈

### DIFF
--- a/src/AbstractAPI/index.js
+++ b/src/AbstractAPI/index.js
@@ -362,15 +362,16 @@ export default class AbstractAPI implements AbstractInterface {
         | CommitDescriptor
         | LayerDescriptor
     ) => {
-      if (objectDescriptor.sha !== "latest") {
+      // Avoid using resolveDescriptor to prevent extra request
+      if (objectDescriptor.sha === "latest") {
+        const commits = await this.commits.list(objectDescriptor);
+        return commits[0];
+      } else {
         const commits = await this.commits.list(
           objectBranchDescriptor(objectDescriptor)
         );
 
         return find(commits, { sha: objectDescriptor.sha });
-      } else {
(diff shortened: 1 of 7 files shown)


src/AbstractAPI/index.js
```diff
@@ -362,15 +362,16 @@ export default class AbstractAPI implements AbstractInterface {
         | CommitDescriptor
         | LayerDescriptor
     ) => {
-      if (objectDescriptor.sha !== undefined) {
```
amccloud: 🍖

src/AbstractCLI/index.js
```diff
@@ -188,15 +188,19 @@ export default class AbstractCLI implements AbstractInterface {
...
-
-      const data = await this.spawn([
-        "commit",
-        objectDescriptor.projectId,
-        objectDescriptor.sha ? objectDescriptor.sha : "latest" // TODO latest
-      ]);
-
-      return data.commit;
+      // Avoid using resolveDescriptor to prevent extra request
+      if (objectDescriptor.sha === "latest") {
```
amccloud: 🍖

src/AbstractAPI/index.js
```diff
@@ -362,15 +362,16 @@ export default class AbstractAPI implements AbstractInterface {
...
+      // Avoid using resolveDescriptor to prevent extra request
+      if (objectDescriptor.sha === "latest") {
+        const commits = await this.commits.list(objectDescriptor);
```
tommoor: We should pass a limit of `1` here, yea? (And the CLI spot further down)

Issue #71, opened 2019-01-09
Add collections.create and collections.update CLI endpoints


bitpshr:
We should add new endpoints to the CLI transport to support the creation of and updating of collections. This will involve new CLI commands to support both operations.

bitpshr:
Closing, as these methods are not currently needed on the CLI transport. See https://github.com/goabstract/abstract-sdk/pull/77#discussion_r250498743 for more information.

PR #74, opened 2019-01-11
feature: add collections.update and collections.create to API transport

### LINKED ISSUE #70: Add collections.create and collections.update API endpoints
We should add new endpoints to the API transport to support the creation of and updating of collections. The required API routes for this already exist, so it should just be a matter of adding these calls to the SDK.

bitpshr:
This pull request adds `collections.update` and `collections.create` endpoints to the API transport.

**Note:** I wasn't totally sure if there was value in exposing these on the API transport since retuned collections are only a subset of what is or can be returned from the CLI. Happy to close this PR if we should table this until full collections can be built using only the API. We still have #71 to handle the CLI side of this.

Resolves #70

### DIFF
--- a/src/AbstractAPI/index.js
+++ b/src/AbstractAPI/index.js
@@ -32,7 +32,8 @@ import type {
   AccessTokenOption,
   AccessToken,
   Activity,
-  Notification
+  Notification,
+  Collection
 } from "../types";
 import randomTraceId from "./randomTraceId";
 import Cursor from "./Cursor";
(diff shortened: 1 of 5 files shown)


tommoor:
> the CLI builds its own supporting file / page / layer data rather than using what the internal API request returns - why is this?

The main reason is so that changes you make locally are immediately reflected, even if offline

bitpshr:
Ah, makes sense @tommoor. In that case, this should be ready for review whenever you or @amccloud have time.

bitpshr:
The upstream PR at https://github.com/goabstract/projects/pull/1288 landed, so this is unblocked for review.

src/AbstractAPI/index.js
```diff
@@ -566,6 +567,35 @@ export default class AbstractAPI implements AbstractInterface {
 
       const data = await unwrapEnvelope(response.json());
       return data.collections[0];
+    },
+    create: async (
+      objectDescriptor: ProjectDescriptor,
+      collection: Collection
```
rickharris: The `Collection` type is what gets returned from the API, not the input for creating a collection. The [params for creating a collection](https://github.com/goabstract/projects/blob/dccecc071648ba37286f2f030cdbf9664c14603c/stable/api/app/controllers/collections_controller.rb#L59-L66) are:

- `name`
- `branchId`
- `description`
- `published`

src/AbstractAPI/index.js
```diff
@@ -566,6 +567,35 @@ export default class AbstractAPI implements AbstractInterface {
...
+          method: "POST",
+          body: collection
+        }
+      );
+
+      return response.json();
+    },
+    update: async (
+      collectionDescriptor: CollectionDescriptor,
+      collection: Collection
```
rickharris: Similar feedback as what I mentioned for the `Collection` type in the `create` action, except for `update`, the params are just

- `name`
- `description`
- `published`

and all are optional.

PR #77, opened 2019-01-16
feature: add collections.update and collections.create to CLI transport

### LINKED ISSUE #71: Add collections.create and collections.update CLI endpoints

bitpshr:
This pull request adds `collections.update` and `collections.create` endpoints to the CLI transport.

Requires https://github.com/goabstract/projects/pull/1288
Resolves #71

### DIFF
--- a/src/AbstractCLI/index.js
+++ b/src/AbstractCLI/index.js
@@ -16,7 +16,8 @@ import type {
   FileDescriptor,
   LayerDescriptor,
   CollectionDescriptor,
-  AccessTokenOption
+  AccessTokenOption,
+  Collection
 } from "../types";
 
 const logSpawn = log.extend("AbstractCLI:spawn");
@@ -316,10 +317,59 @@ export default class AbstractCLI implements AbstractInterface {
     },
     info: (collectionDescriptor: CollectionDescriptor) => {
       return this.spawn([
-        "collection",
+        "collection load",
         collectionDescriptor.projectId,
         collectionDescriptor.collectionId
       ]);
+    },
+    create: async (
+      projectDescriptor: ProjectDescriptor,
+      collection: Collection
+    ) => {
+      const name = collection.name ? ["--name", collection.name] : [];
+      const branchId = collection.branchId
+        ? ["--branchID", collection.branchId]
+        : [];
+      const description = collection.description
+        ? ["--description", collection.description]
+        : [];
+      const published = collection.published
+        ? ["--published", String(collection.published)]
+        : [];
+
+      const data = await this.spawn([
+        "collection",
+        "create",
+        projectDescriptor.projectId,
+        ...name,
+        ...branchId,
+        ...description,
+        ...published
+      ]);
+      return data.collection;
+    },
+    update: async (
+      collectionDescriptor: CollectionDescriptor,
+      collection: Collection
+    ) => {
+      const name = collection.name ? ["--name", collection.name] : [];
+      const description = collection.description
+        ? ["--description", collection.description]
+        : [];
+      const published = collection.published
+        ? ["--published", String(collection.published)]
+        : [];
+
+      const data = await this.spawn([
+        "collection",
+        "update",
+        collectionDescriptor.projectId,
+        collectionDescriptor.collectionId,
+        ...name,
+        ...description,
+        ...published
+      ]);
+      return data.collection;
     }
   };
 
--- a/src/AbstractCLI/index.js
+++ b/src/AbstractCLI/index.js
@@ -391,8 +391,8 @@ export default class AbstractCLI implements AbstractInterface {
       return await this.spawn([
         "branch",
         "load",
-        branchDescriptor.branchId,
-        branchDescriptor.projectId
+        branchDescriptor.projectId,
+        branchDescriptor.branchId
       ]);
     }
   };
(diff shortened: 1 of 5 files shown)


bitpshr:
This PR has been rebased and updated to include a necessary `peerDependency` upgrade to `abstract-cli@1.0.0`, which is now live. This (and the [API transport PR](https://github.com/goabstract/abstract-sdk/pull/74)) should be ready for a review whenever you have a chance @tommoor and @amccloud 🙏

src/AbstractCLI/index.js
```diff
@@ -316,10 +317,59 @@ export default class AbstractCLI implements AbstractInterface {
...
+      const data = await this.spawn([
+        "collection",
+        "create",
+        projectDescriptor.projectId,
+        ...name,
+        ...branchId,
+        ...description,
+        ...published
+      ]);
+      return data.collection;
```
rickharris: So, there's kind of a weird thing happening here using the CLI transport for collections.

The CLI transport:
- Uses the API internally to create the collection
- Uses the local repo to build up all the file data (files, pages, layers) needed to render the collection
- Returns the collection & relevant policy (comes from the API) along with the files, pages and layers (comes from local repo)

And then here, we throw out the policies, files, pages, and layers and return only the collection object that originally came from the API.

So, is there any real point in having this be part of the CLI transport if:

- Collections are already considered an online-only feature
- We're not providing anymore more than using the API transport would give us

rickharris: What I said above _technically_ doesn't apply to `create`, since as of right now you can't technically create a collection with collection layers already in it, so the CLI command that populates files, pages, layers is basically a noop, but it does apply to `update`. And could possibly apply to `create` if we ever let you create collection layers as part of the collection creation process.

tommoor: I think there were a couple of reasons for doing this, unfortunately the [original PR](https://github.com/goabstract/JIRA/issues/3564) doesn't provide much color.

Two things I can think of that would need considering:

- It's nice that when you commit offline the collection updates in real time
- Although we're not doing it now there is potential here to reduce load on our servers by not requesting all the file information, I believe this was the original intent.

rickharris: I fully agree with both of those points, especially with regard to using `abstract-cli` directly in desktop. However, as this `abstract-sdk` PR stands, we won't get either of those benefits, as we're only returning the collection itself and not the associated file information from the repo.

> It's nice that when you commit offline the collection updates in real time

this will only happen for auto-updating collection layers, which need the extra file info from the cli

> there is potential here to reduce load on our servers by not requesting all the file information

this transport as it stands now will reduce server load, but we won't get any of the updated file information by only returning `data.collection`.

bitpshr: One option is to modify the signatures in both transports to return the `collection` along with its data, rather than just the collection itself. It was probably an oversight on my part not to return this data in the first place.

rickharris: I think I mis-spoke before when I said

> What I said above actually doesn't apply to create ... but it does apply to update

The API doesn't actually return file data for create or update. It only returns file data (if you ask it to) on `CollectionsController#[:show, :index]`

This means that `abstract-cli` is technically returning more data than `api` for the collection update action, which I believe leaves us with two options:

1. Add the ability to return file data on `CollectionsController#update` to match cli
2. Stop returning file data in `abstract-cli collection update`

However, if we do the latter, then the CLI transport layer for collections becomes useless, because all we're doing is shelling out to abstract-cli to make an API request and do some unnecessary JSON parsing. We've lived this long without needing to make collection updates return file data, and updating the collection via the API shouldn't technically change any of its `collectionLayers` data, so maybe this is the way to go?

bitpshr: +1 for updating the CLI in this case to return the same data (collection only, no file data) as the API since this has been sufficient for downstream apps so far. cc @tommoor; this is what I was noticing as well re: differences between API and CLI data completeness.

bitpshr: I just noticed the CLI and API transports are returning different data for existing collection endpoints on `master`, so this discrepancy isn't necessarily specific to `update` at the SDK level.

Should the CLI and API transports always return the same data? If they should, does this mean our CLI transport should throw out the extra file data before resolving each endpoint?

rickharris: > I just noticed the CLI and API transports are already returning different data

that sounds bad to me since you're not really supposed to know what transport you're using. cc @amccloud

rickharris: > I just noticed the CLI and API transports are already returning different data for existing collection endpoints on master

apart from this, are we in agreement that the CLI transport doesn't even need `create` and `update` methods for collections? Since all the CLI is doing is internally using the API?

bitpshr: Thanks for helping me understand this @rickharris. I'll go ahead and close this PR. I opened up https://github.com/goabstract/abstract-sdk/issues/79 to consolidate returned data between the CLI and API for `info` and `list` collection methods.

bitpshr:
These methods aren't actually needed since the CLI calls out the API under the hood (briefly asked about [here](https://github.com/goabstract/abstract-sdk/pull/74#issue-244091597).)

PR #80, opened 2019-01-24
bugfix: consolidate API and CLI collection data

### LINKED ISSUE #79: API and CLI transports return different collection data
The API and CLI transports currently return different data when querying collections. The CLI returns the collection along with its file data, whereas the API only returns the collection.

bitpshr:
This pull request fixes an issue where the CLI would return different data than the API for `collections` methods. The CLI was returning collections along with file data, so to maintain parity, the API also returns file data as well (though the file data will only be populated if the user opts-in to avoid a performance hit.)

Fixes #79

### DIFF
--- a/src/types.js
+++ b/src/types.js
@@ -551,6 +551,18 @@ export type CollectionLayer = {
   order: number
 };
 
+export type File = {
+  applicationDocumentVersion: number,
+  applicationVersion: string,
+  id: string,
+  isLibrary: boolean,
+  lastChangedAtSha: string,
+  name: string,
+  projectId: string,
+  sha: string,
+  type: string
+};
+
 export type Collection = {
   id: string,
   userId: string,
@@ -1389,6 +1401,13 @@ export type CursorResponse<T> = {
   meta: CursorMeta
 };
 
+export type CollectionPage = {
+  collections: Collection[],
+  files: File[],
+  pages: Page[],
+  layers: Layer[]
+};
+
 export type AccessToken = ?string | ShareDescriptor;
 export type AccessTokenOption =
   | AccessToken // TODO: Deprecate?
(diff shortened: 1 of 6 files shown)


rickharris:
I think we'd probably need to fix this the opposite way. For us to use this in our own product, we need file data for collections.

You can still request file data from the API & it'll get that from gitapi. It's what we do for web. You can tell the API, for each collection, [how many layers you'd like data for](https://github.com/goabstract/ui/blob/f6edbeeebea83a69e38e60c36731df1504e39ba5/web/src/api/index.js#L808-L809).

This is useful for `collections.index`, so that you can just request a few layers for each collection, enough to show some kind of preview, but not enough to bog the server down. And for `collections.show`, you can pass `all` to return all file data for the collection.

src/types.js
```diff
@@ -551,6 +551,18 @@ export type CollectionLayer = {
...
+  applicationDocumentVersion: number,
+  applicationVersion: string,
+  id: string,
+  isLibrary: boolean,
+  lastChangedAtSha: string,
+  name: string,
+  projectId: string,
+  sha: string,
+  type: string
+};
```
bitpshr: We were inadvertently using the built-in `File` type and never actually defined our own.

bitpshr:
Note: The CLI was returning collections along with file data, so to maintain parity, the API now returns file data as well with this PR. Note that the API will return empty arrays for `files`, `pages`, and `layers` when hitting `collections.list` unless a user explicitly passes `layersPerCollection`. This is to prevent a performance hit by default, while still maintaining parity with the CLI.

This should be ready for a review whenever there's time.

cc @tommoor @amccloud

bitpshr:
Thanks for the review @amccloud.

Good point about splitting this back out into multiple calls. Just making sure I understand correctly: What type would we return from `collections.list` and `collection.info` from each transport? `Collection`, and then the consumer would be responsible for querying additional file / page / layer data, which wouldn't be a big deal since it'd be cached?

PR #81, opened 2019-01-24
Upgrade flow@90

amccloud:
Did this just to test and was surprised it just worked. Not ready to merge yet. We wouldn't want to accidentally adopt a new flow feature that can't be used downstream


bitpshr:
Closing this to keep the PR queue clean for now @amccloud; I figured we could revisit this when the UI is ready to move to `flow@0.90.0` as well.

PR #82, opened 2019-01-24
feature: add custom error handling

### LINKED ISSUE #56: Add rate limit handling
I'd suggest the following but am of course open to other things:

- First pass: Add a custom error that is thrown when the rate limit is hit
- Second pass: Internally handle the rate limit and wait for the appropriate amount of time based on the response headers before continuing.

bitpshr:
This pull request updates the API transport by adding a set of custom errors along with a convenience function to analyze `Response` objects and conditionally throw the correct error.

Resolves #56
Resolves #82


src/AbstractAPI/errors.js
```diff
@@ -0,0 +1,96 @@
...
+type ErrorData = {|
+  url: string,
+  body: mixed
+|};
+
+class BaseError extends Error {
+  constructor(message: string) {
+    super(message);
+    this.name = this.constructor.name;
+    Error.captureStackTrace(this, this.constructor);
```
bitpshr: This is a Node API that helps stack trace readability: From the [`captureStackTrace` documentation](https://nodejs.org/api/errors.html#errors_error_capturestacktrace_targetobject_constructoropt):

> The optional `constructorOpt` argument accepts a function. If given, all frames above `constructorOpt`, including `constructorOpt`, will be omitted from the generated stack trace. The `constructorOpt` argument is useful for hiding implementation details of error generation from an end user.

src/AbstractAPI/errors.js
```diff
@@ -0,0 +1,96 @@
...
+  }
+}
+
+export class InternalServerError extends BaseError {
+  data: ErrorData;
+
+  constructor(url: string, body: mixed) {
+    super("Internal server error.");
+    this.data = { url, body };
```
bitpshr: In practice, it was useful to include (at least) the request URL and any request body with each error.

src/AbstractAPI/errors.js
```diff
@@ -0,0 +1,94 @@
...
+  constructor(url: string, body: mixed) {
+    super("Not found.");
+    this.data = { url, body };
+  }
+}
+
+export class RateLimitError extends BaseError {
+  data: {
+    ...ErrorData,
+    resetsAt: ?number
```
bitpshr: This error class includes additional error data specific to rate limiting. The API returns an ambiguously-named `resets` response header that's an epoch timestamp indicating when requests can be made again. For clarity, we expose this as `resetsAt` so it's clear this is not how long a user must wait before making additional requests again ([which is technically what should be returned](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After) with `429` errors.)

src/AbstractAPI/index.js
```diff
@@ -33,15 +33,14 @@ import type {
   AccessToken,
   Activity,
   Notification,
-  Collection,
```
bitpshr: Cleanup; this wasn't used.

src/AbstractAPI/index.js
```diff
@@ -108,7 +107,7 @@ export default class AbstractAPI implements AbstractInterface {
     }
   }
 
-  async fetch(input: string | URL, init: Object = {}, hostname?: ?string) {
+  async fetch(input: string, init: Object = {}, hostname?: ?string) {
```
bitpshr: The full signature for `fetch` should be:

```js
fetch(input: RequestInfo, init?: RequestInit): Promise<Response>;
```

We could modify this local `fetch` signature to match the global `fetch` signature above, but since we're in control of this local function and how it's used, it makes sense to limit to only accept what we know we'll give it. For now, that's a `string` for its URL and an `object` for its initialization data (both of which satisfy `RequestInfo` and `RequestInit`, respectively.)

src/AbstractAPI/errors.js
amccloud: I wonder if we'll have similar errors for cli. The seems like an error type that could come from cli. In that case we might not have a url unless we fake it

src/AbstractAPI/errors.js
```diff
@@ -0,0 +1,94 @@
...
+  };
+
+  constructor(url: string, body: mixed, response: Response) {
+    const resetSeconds = response.headers.get("ratelimit-reset");
+    const resetsAt = Number(resetSeconds && Number(resetSeconds) * 1000);
```
amccloud: is outer `Number` needed?

src/AbstractAPI/errors.js
bitpshr: Good point. I went ahead and took care of CLI errors in this PR as well now, as it made sense to see how errors could be shared between the two transports. I ended up generalizing `url` to `path`, so it could be either a fetch URL or the CLI path, and `body` is either the fetch body or CLI input args. I figured we could tweak this error data depending on what users end up needing or finding important.

src/AbstractAPI/errors.js
bitpshr: No, temporary insanity on my part :/ Fixed!

bitpshr:
Thanks for the review @amccloud. Good call on thinking about CLI error handling; I went ahead and took care of that as well. Ready for another pass when you have a chance.

src/AbstractCLI/index.js
```diff
@@ -125,8 +125,13 @@ export default class AbstractCLI implements AbstractInterface {
...
         if (errorCode !== 0) {
-          logError(stderrBuffer.toString());
-          reject(stderrBuffer); // Reject stderr for non-zero error codes
+          const response = JSON.parse(stderrBuffer.toString());
+          catchCLIError(
+            response,
+            spawnArgs[0],
+            { ...(spawnArgs[1]: any) },
+            reject
+          );
```
amccloud: maybe `throwXError` since it's the source of the exception?

I wonder if we could avoid the `reject` argument and can keep the throwing behavior the same. While redundant would this work?


```js
try {
  throwCLIError(/* ... */)
} catch (error) {
  reject(error);
}
```

bitpshr: Great call, much nicer. Updated all around.

bitpshr:
@amccloud good catch with the `response.clone` issue; should be fixed now. Thanks for the reviews so far.

src/AbstractCLI/index.js
```diff
@@ -125,8 +125,12 @@ export default class AbstractCLI implements AbstractInterface {
         logClose(errorCode);
 
         if (errorCode !== 0) {
-          logError(stderrBuffer.toString());
-          reject(stderrBuffer); // Reject stderr for non-zero error codes
+          const response = JSON.parse(stderrBuffer.toString());
```
tommoor: There's a decent chance this JSON parse throws too with it being the result of concatenating various stdout data chunks. Might want to move inside it's own catch block

bitpshr: The CLI actually helps us out here: any errors it returns are always formatted as stringified JSON (even when unexpected `panic` calls), so we can safely parse both known and unknown errors as a result.

PR #89, opened 2019-01-31
Add flow-typed modules

amccloud:
`0.7.0` of the SDK is [throwing flow errors downstream](https://circleci.com/gh/goabstract/ui/9248) because we have libdefs for things like lodash downstream that are missing in the SDK. When running flow in a downstream project the libdefs from that project are used to check the SDK

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,5 +1,5 @@
 [ignore]
-./lib/.*
+<PROJECT_ROOT>/lib/.*
 
 [include]
 
--- a/flow-typed/npm/query-string_v6.x.x.js
+++ b/flow-typed/npm/query-string_v6.x.x.js
@@ -1,0 +1,27 @@
+// flow-typed signature: f837533c0e35d92743f6b45571100838
+// flow-typed version: 0671bf6d89/query-string_v6.x.x/flow_>=v0.32.x
+
+declare module 'query-string' {
+  declare type ArrayFormat = 'none' | 'bracket' | 'index'
+  declare type ParseOptions = {|
+    arrayFormat?: ArrayFormat,
+  |}
+
+  declare type StringifyOptions = {|
+    arrayFormat?: ArrayFormat,
+    encode?: boolean,
+    strict?: boolean,
+    sort?: false | <A, B>(A, B) => number,
+  |}
+
+  declare type QueryParameters = {
+    [string]: string | Array<string> | null
+  }
+
+  declare module.exports: {
+    extract(str: string): string,
+    parse(str: string, opts?: ParseOptions): QueryParameters,
+    parseUrl(str: string, opts?: ParseOptions): { url: string, query: QueryParameters },
+    stringify(obj: QueryParameters, opts?: StringifyOptions): string,
+  }
+}
--- a/flow-typed/npm/query-string_v6.x.x.js
+++ b/flow-typed/npm/query-string_v6.x.x.js
@@ -1,5 +1,5 @@
-// flow-typed signature: f837533c0e35d92743f6b45571100838
-// flow-typed version: 0671bf6d89/query-string_v6.x.x/flow_>=v0.32.x
+// flow-typed signature: 9096718d5c8abc7d9ae9aadbe0ce565a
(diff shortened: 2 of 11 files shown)


flow-typed/npm/query-string_v6.x.x.js
```diff
@@ -0,0 +1,33 @@
...
+
+  declare type QueryParameters = {
+    [string]: string | Array<string> | null
+  }
+
+  declare module.exports: {
+    extract(str: string): string,
+    parse(str: string, opts?: ParseOptions): QueryParameters,
+    parseUrl(str: string, opts?: ParseOptions): { url: string, query: QueryParameters },
+    stringify(obj: ObjectParameters, opts?: StringifyOptions): string,
```
amccloud: This was modified and is not the original query-string libdef.

Patch submitted upstream https://github.com/flow-typed/flow-typed/pull/3114

.flowconfig
```diff
@@ -1,5 +1,5 @@
 [ignore]
-./lib/.*
+<PROJECT_ROOT>/lib/.*
```
amccloud: I still noticed lib files in flow errors without this

flow-typed/npm/query-string_v6.x.x.js
rickharris: Looks like it’s merged now 🙌

src/AbstractAPI/test.js
```diff
@@ -345,7 +350,7 @@ describe("AbstractAPI", () => {
       [
         "files.info",
         buildFileDescriptor({ sha: "latest" }),
-        { responses: [responses.commits.list()] }
+        { responses: [responses.commits.list(), responses.files.list()] }
```
amccloud: The first call to list is to resolve latest and then the second call is to list the commits we call find on

src/AbstractAPI/index.js
```diff
@@ -397,7 +397,10 @@ export default class AbstractAPI implements AbstractInterface {
           { limit: 1 }
         );
 
-        return find(commits, { sha: objectDescriptor.sha });
+        const commit = find(commits, { sha: objectDescriptor.sha });
+        if (!commit) throw new Error("Commit not found");
```
tommoor: It'd be nice to see a custom error for these 404 equivalents, maybe for a future PR

amccloud: I went to use the existing errors but realized we made them a bit too specific. 

Here we don't have access to url to throw any of the existing exceptions  correctly. We need something like `new NotFoundError({ url, body })` instead

bitpshr: Why don't we have access to the URL? Can't we build it the same way `fetch` does using `this.apiUrl` and throw it manually? We could also just use "commits" for now; nothing in the error code mandates that `path` has to be a real URL:

```js
throw new NotFoundError("commits", { sha: objectDescriptor.sha });
```

bitpshr:
> The flow errors in the sdk existed because we were missing type definitions for lodash/find/etc.

I'm mostly wondering why we include external library definitions in our repository at all. Can they be installed at install time or build time via the `flow-typed` CLI? What's the advantage of committing these?

amccloud:
@bitpshr with that you lose the ability to modify definitions like I did in this PR. Also we’d likely want to add definitions that do not exist on flow-typed

My change to flow-typed was merged quickly but that’s not always the case

bitpshr:
Ah, ok, thanks for this. I gotta say, it's pretty brutal that we have to include installable typings in our version control only so we can pave over errors in the provided community typings. (_Insert some pro-TypeScript comment about the accuracy and automatic nature of third-party TS typings. 😅_)

This PR still looks good to me though, thanks for fixing this.

src/AbstractAPI/index.js
amccloud: @bitpshr maybe a private `commits._url()` method that fetch and the error can use? Still sorta misleading because this is a virtual lookup. The url could `200` but we would still raise `404` for the url

Any reason we should not make the error arguments optional?

bitpshr: I think we totally should. I will update.

Issue #92, opened 2019-02-08
previews: Add method to retrieve preview url


tommoor:
`previews.raw` is great if you want a copy of the file to save in your own datastore, but it's an awkward interface for displaying in the web browser.

Lets add an additional method to grab a data-uri for the image using the `bufferToUrl` helper we have internally. This issue was triggered by confusion from a customer attempting to use the `webUrl` method…

ServusJon:
Is there a method to get the url to artboard for the browser? So you can use the Inspect functionality?

tommoor:
That would be `previews.webUrl` which links to the artboard in our web app.

ServusJon:
We tried that link but it always showed us "forbidden" .
Also, is there a generalized link to an artboard that isn't connected to a commit but to the most recent version of it?

amccloud:
@ServusJon whoops! `webUrl` is actually pointing to the preview image file but that requires authentication. 

PR incoming to get direct access to the image and to update webUrl. In the meantime you could try replacing `previews.goabstract.com` with `app.goabstract.com` in the current webUrl

ServusJon:
Cool! That will work!

amccloud:
`previews.url` was added in `v0.7.5` https://sdk.goabstract.com/docs/abstract-api/#retrieve-a-preview-image-url

PR #94, opened 2019-02-14
Add previews.url

amccloud:
For https://github.com/goabstract/abstract-sdk/issues/92

- Add `previews.url`
- Update `Preview.webUrl` to point to app.goabstract.com

### Todo

- [x] Test coverage

### DIFF
--- a/docs/abstract-api.md
+++ b/docs/abstract-api.md
@@ -871,7 +871,7 @@ fs.writeFile(`preview.png`, Buffer.from(arrayBuffer), (err) => {
 });
 ```
 
-### Retrieve an image url
+### Retrieve a preview image url
 
 > Note: The `previews.url` method requires an environment with [URL.createObjectURL](https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL). If you are using node, you will need to save the image to a file with [`previews.raw`](#retrieve-an-image-file)
 
@@ -880,7 +880,7 @@ fs.writeFile(`preview.png`, Buffer.from(arrayBuffer), (err) => {
 > Get an image as a _temporary_ blob url which can be displayed directly in an image tag or downloaded. The url exists only as long as the current browser session and should not be saved to a database directly.
 
 ```js
-abstract.previews.info({
+abstract.previews.url({
   projectId: "616daa90-1736-11e8-b8b0-8d1fec7aef78",
   branchId: "master",
   fileId: "51DE7CD1-ECDC-473C-B30E-62AE913743B7",
--- a/docs/abstract-api.md
+++ b/docs/abstract-api.md
@@ -875,9 +875,9 @@ fs.writeFile(`preview.png`, Buffer.from(arrayBuffer), (err) => {
 
 > Note: The `previews.url` method requires an environment with [URL.createObjectURL](https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL). If you are using node, you will need to save the image to a file with [`previews.raw`](#retrieve-an-image-file)
 
-`previews.url(LayerDescriptor): Promise<Preview>`
+`previews.url(LayerDescriptor): Promise<string>`
 
-Load a url object for the preview
+> Get an image as a _temporary_ blob url which can be displayed directly in an image tag or downloaded. The url exists only as long as the current browser session and should not be saved to a database directly.
 
 ```js
 abstract.previews.info({
--- a/docs/abstract-api.md
+++ b/docs/abstract-api.md
@@ -871,6 +871,25 @@ fs.writeFile(`preview.png`, Buffer.from(arrayBuffer), (err) => {
 });
 ```
 
+### Retrieve an image url
+
+> Note: The `previews.url` method requires an environments with [URL.createObjectURL](https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL). If you are using node, you will need to save the image to a file with [`previews.raw`](#retrieve-an-image-file)
+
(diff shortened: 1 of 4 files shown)


docs/abstract-api.md
```diff
@@ -871,6 +871,25 @@ fs.writeFile(`preview.png`, Buffer.from(arrayBuffer), (err) => {
 });
 ```
 
+### Retrieve an image url
+
+> Note: The `previews.url` method requires an environment with [URL.createObjectURL](https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL). If you are using node, you will need to save the image to a file with [`previews.raw`](#retrieve-an-image-file)
+
+`previews.url(LayerDescriptor): Promise<Preview>`
```
tommoor: Response type is incorrect here

src/AbstractAPI/index.js
```diff
@@ -537,19 +541,31 @@ export default class AbstractAPI implements AbstractInterface {
...
-        options
+        `projects/${layerDescriptor.projectId}/commits/${layerDescriptor.sha}/files/${layerDescriptor.fileId}/layers/${layerDescriptor.layerId}`
       );
 
       return response.arrayBuffer();
+    },
+    url: async (layerDescriptor: LayerDescriptor) => {
+      if (!Blob || !URL || !DataView) {
+        throw new Error(
+          `"previews.url" requires an environment with Blob, URL and DataView`
```
tommoor: We can know if they are running in node to some degree of certainty by checking if `process` exists, would be nice to show an EVEN MORE useful message here and point them to the raw method as you have in the docs above…

docs/abstract-api.md
```diff
@@ -871,6 +871,25 @@ fs.writeFile(`preview.png`, Buffer.from(arrayBuffer), (err) => {
...
+
+Load a url object for the preview
```
tommoor: I think it's worth being really unambiguous here as most consumers of this API will probably expect a http url. Something like:

> Get an image as a _temporary_ blob url which can be displayed directly in an image tag or downloaded. The url exists only as long as the current browser session and should not be saved to a database directly.

amccloud:
@tommoor ready for another 👀

tommoor:
👍 to the updates, thanks!

docs/abstract-api.md
```diff
@@ -871,6 +871,25 @@ fs.writeFile(`preview.png`, Buffer.from(arrayBuffer), (err) => {
 });
 ```
 
+### Retrieve an image url
```
bitpshr:
```suggestion
### Retrieve a preview image url
```

docs/abstract-api.md
```diff
@@ -871,6 +871,25 @@ fs.writeFile(`preview.png`, Buffer.from(arrayBuffer), (err) => {
...
+
+> Note: The `previews.url` method requires an environment with [URL.createObjectURL](https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL). If you are using node, you will need to save the image to a file with [`previews.raw`](#retrieve-an-image-file)
+
+`previews.url(LayerDescriptor): Promise<string>`
+
+> Get an image as a _temporary_ blob url which can be displayed directly in an image tag or downloaded. The url exists only as long as the current browser session and should not be saved to a database directly.
```
bitpshr: 👍

docs/abstract-api.md
```diff
@@ -871,6 +871,25 @@ fs.writeFile(`preview.png`, Buffer.from(arrayBuffer), (err) => {
...
+
+```js
+abstract.previews.info({
```
bitpshr:
```suggestion
abstract.previews.url({
```

src/AbstractAPI/index.js
```diff
@@ -537,19 +541,31 @@ export default class AbstractAPI implements AbstractInterface {
...
-        `projects/${layerDescriptor.projectId}/commits/${layerDescriptor.sha}/files/${layerDescriptor.fileId}/layers/${layerDescriptor.layerId}`,
-        options
+        `projects/${layerDescriptor.projectId}/commits/${layerDescriptor.sha}/files/${layerDescriptor.fileId}/layers/${layerDescriptor.layerId}`
       );
 
       return response.arrayBuffer();
+    },
+    url: async (layerDescriptor: LayerDescriptor) => {
+      if (!Blob || !URL || !DataView) {
+        throw new Error(
```
bitpshr: Is it worth creating a custom error for this for consistency?

src/AbstractAPI/index.js
```diff
@@ -50,7 +50,8 @@ const logFetch = log.extend("AbstractAPI:fetch");
 export type Options = {
   accessToken: AccessTokenOption,
   apiUrl?: string,
-  previewsUrl?: string
+  previewsUrl?: string,
+  webUrl?: string
```
bitpshr: Is it really not safe to derive the `webUrl` from the `apiUrl`? Both are in our control and it would reduce our option footprint. Maybe even for `previewUrl` too.

amccloud: Early on I thought about supplying just a hostname and have everything added around it. Doesn't play well with our development environments which use multiple ports and paths for these services https://github.com/goabstract/ui/blob/master/shared/src/components/Embed/connector.js#L39

src/AbstractAPI/index.js
amccloud: I think it makes sense to only have custom errors for things we document and expect people to handle and possibly recover from. Should this be documented that way?

Issue #41, opened 2018-11-25
Add asynchronous iteration support


bitpshr:
We should add asynchronous iteration support for paginated list methods using an external function that leverages the cursor-based pagination we have in place.

PR #98, opened 2019-02-20
feature: SDK reorganization

bitpshr:
**Note: Ignore this diff, it's pretty useless. Read the description instead.**

## Intro

@amccloud  and I spoke in SF about pain points around external SDK integration and internal SDK maintenance that have been surfaced by heavier internal usage. @rickharris has also mentioned usage difficulties as an external consumer in a Flow-typed project. A [new SDK branch](https://github.com/goabstract/abstract-sdk/tree/feature/types-and-modules/) attempts to reorganize existing SDK logic into a modular structure that reduces type maintenance, fixes downstream type consumption, and allows for runtime-determined method implementations. Most insight can be gained by checking out the [new directory structure](https://github.com/goabstract/abstract-sdk/tree/feature/types-and-modules/src).

**TLDR;** You can verify this now by installing `abstract-sdk-dev` from NPM. Usage should be the same except for client instantiation. Ping me if you have any issues.

## Motivation

Now that the SDK has been live and put to use in some form both internally and externally, areas for improvement have been identified, including duplicated types and interfaces, triplicated method signatures, method existence checks due to transport interface variability, a less-than-ideal `AUTO` transport, and two 1000+ line transport implementations that mostly consist of duplicated method signatures (that are already defined in a separately maintained transport interface.)

As the SDK approaches maturity as a usable, production-ready tool, its structure should be revisited to to make internal development easier, safer, and more predictable, and to strengthen type-safe integration with consumers. By reorganizing existing logic into a new structure with strong typing in mind, several immediate advantages can be realized:

## Advantages

- **The public SDK interface is defined by the code, not a manually-maintained interface.**
- **Public method signatures are not duplicated between transports.**
- **Downstream type consumption works as-expected without issue.**
- **Flow-types are installed as part of NPM installs and are .gitignored.**
- Common code, like input validation, is not duplicated between transports.
- Each endpoint lives in its own module extending a common base class.
- Every applicable endpoint supports `sha: "latest`.
- The `Client` can be a class again since it doesn't return transport instances.
- API vs. CLI implementations can be chosen at runtime, based on factors like connectivity.
- Informative errors are surfaced if a method isn't supported for a given transport.
- All public and private APIs are exposed with total type safety.
- Removed dependance on `JSON-stream`, `lodash`, and legacy `babel-*` NPM packages.

Most importantly, adding a new endpoint no longer will require 1) adding a method signature to the manually-maintained `AbstractSDK` interface, 2) adding an identical signature with implementation to the SDK transport, then 3) adding a third and identical signature with implementation to the CLI transport. Instead, adding a new endpoint will require 1) adding a new module with the method signature you need.

## Changes

This was mostly a reorganization of existing logic. Below is a summary of all relevant changes:

1. The manually-maintained [`AbstractSDK` interface went away](https://github.com/goabstract/abstract-sdk/pull/98/files#diff-3a699a0719518f967e9b2df3f5549e02L1417).
1. Each endpoint is a module, like [`Branches`](https://github.com/goabstract/abstract-sdk/blob/feature/types-and-modules/src/endpoints/Branches.js).
1. Each endpoint is tested, like [`Branches.test`](https://github.com/goabstract/abstract-sdk/blob/feature/types-and-modules/src/endpoints/Branches.test.js).
1. Each endpoint module extends [`BaseEndpoint`](https://github.com/goabstract/abstract-sdk/blob/feature/types-and-modules/src/endpoints/BaseEndpoint.js).
1. `BaseEndpoint` provides common methods: [`apiRequest`](https://github.com/goabstract/abstract-sdk/blob/feature/types-and-modules/src/endpoints/BaseEndpoint.js#L84-L93), [`apiRawRequest`](https://github.com/goabstract/abstract-sdk/blob/feature/types-and-modules/src/endpoints/BaseEndpoint.js#L95-L104), [`cliRequest`](https://github.com/goabstract/abstract-sdk/blob/feature/types-and-modules/src/endpoints/BaseEndpoint.js#L106-L153), etc.
1. Transport implementation is chosen [granularly and at runtime](https://github.com/goabstract/abstract-sdk/blob/feature/types-and-modules/src/endpoints/BaseEndpoint.js#L56-L82).
1. The [`Client`](https://github.com/goabstract/abstract-sdk/blob/feature/types-and-modules/src/Client.js) is a class that validates CLI options and instantiates endpoints.
1. `sha` was removed from `BranchDescriptor`, and relevant uses were updated to correctly use `CommitDescriptor`. This error was causing any endpoint expecting a `BranchDescriptor` to also include an `sha` when called.
1. An external [`paginate`](https://github.com/goabstract/abstract-sdk/blob/8fccaeae74ff9f74925a46eaa29cf970b71da057/src/paginate.js) function was added to support `await-for-of` asynchronous iteration.

## Summary

From a high level, this approach sets out to modularize the ever-growing transport implementations in a way that removes the need for duplicated type signatures throughout our codebase. Since API and CLI implementations are defined alongside one another, inside the same parent method signature, Flow guarantees they adhere to the same interface. Still further, Flow uses these concrete method signatures as the `AbstractSDK` interface expected by consumers, so nothing has to be maintained.

Resolves https://github.com/goabstract/abstract-sdk/issues/99
Resolves https://github.com/goabstract/abstract-sdk/issues/41


.babelrc
```diff
@@ -1,6 +1,5 @@
 {
   "presets": [
-    "@babel/preset-react",
```
bitpshr: This was not needed for anything.

bitpshr: This was not needed for anything.

.editorconfig
```diff
@@ -0,0 +1,13 @@
+# http://editorconfig.org
```
bitpshr: Added for sanity; I use `editorconfig` so this helps keep my editor straight since I'm a hard-tab type of person normally in other projects.

.flowconfig
```diff
@@ -4,7 +4,6 @@
 [include]
 
 [libs]
-definitions
```
bitpshr: We have no `definitions` defined anymore. See comment below re: globals.

.gitignore
```diff
@@ -1,11 +1,12 @@
 .DS_Store
-node_modules
```
bitpshr: Mostly just alphabetization.

.npmignore
```diff
@@ -1,5 +0,0 @@
-src
```
bitpshr: Since we include so many fewer files than we exclude, I went with `files` in `package.json` instead of `.npmignore`.

definitions/globals.js
```diff
@@ -1,8 +0,0 @@
-// @flow
-
-declare type PROSketchBootstrapT = {
-  documentKey: (*) => string
-};
-
-declare var NSString: Object;
-declare var PROSketchBootstrap: PROSketchBootstrapT;
```
bitpshr: Sketch-specific declarations were moved into the file they were used in, which also fixed downstream typing issues. Nothing in downstream projects ever pulled in these types from `definitions/globals.js`, so downstream project compilers never knew about them.

jest.config.js
```diff
@@ -1,5 +1,3 @@
 module.exports = {
-  testPathIgnorePatterns: ["<rootDir>/lib"],
-  snapshotSerializers: ["jest-serializer-path"],
-  modulePathIgnorePatterns: ["<rootDir>/.*/__mocks__"] // https://github.com/facebook/jest/issues/2070
+  testPathIgnorePatterns: ["<rootDir>/lib"]
```
bitpshr: Tests were kept straightforward and modular: we are not using snapshots, and instead mock on a per-request basis. We simply compare expected return values. This takes a lot of the configuration magic out of our testing setup.

scripts/install-types.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
```
bitpshr: This is a script run as a `postinstall` NPM script that automatically installs required `flow-typed` definitions. This allows us to `.gitignore` the definitions from our repository, and no issues with the current typings have warranted keeping them in our source tree...yet.

src/Client.js
```diff
@@ -0,0 +1,105 @@
...
+  data: Data;
+  files: Files;
+  layers: Layers;
+  notifications: Notifications;
+  organizations: Organizations;
+  pages: Pages;
+  previews: Previews;
+  projects: Projects;
+  shares: Shares;
+  users: Users;
```
bitpshr: This basically defines the formerly-manually-maintained `AbstractSDK` interface through the code itself. The `Client` is the `AbstractSDK` interface. Rather than making sure the shape of our code adheres to an interface, the shape of our code defines that interface.

jest.config.js
amccloud: Do we lose coverage with the change? Most methods don't have return values worth asserting but it is worth asserting that we are generating the correct effect to send to api or cli. Can we assert that fetch/cli is called with the correct arguments with this approach?

package.json
```diff
@@ -1,23 +1,22 @@
...
-  },
   "scripts": {
-    "prepublish": "yarn run build",
-    "build": "babel ./src --ignore **/*/__mocks__,**/test.js -d ./lib",
-    "prebuild": "flow-copy-source --ignore '**/*/__mocks__/*.js' --ignore '**/*test.js' ./src ./lib",
-    "lint": "eslint .",
+    "postinstall": "./scripts/install-types.sh",
+    "lint": "eslint . --ignore-path .gitignore",
     "test": "jest",
-    "test-inspect": "node --inspect-brk ./node_modules/.bin/jest --runInBand",
```
amccloud: Why remove test-inspect? It's pretty useful and somewhat hard to recall

src/endpoints/BaseEndpoint.js
```diff
@@ -0,0 +1,208 @@
...
+const logCLIRequest = log.extend("AbstractCLI:request");
+const logCLIResponse = log.extend("AbstractCLI:response");
+const minorVersion = version.split(".", 2).join(".");
+
+export type EndpointHandler<T> = {
+  api?: () => T,
+  cli?: () => T
+};
+
+export default class BaseEndpoint {
```
amccloud: Thoughts on just calling this `Endpoint`?

src/endpoints/BaseEndpoint.js
```diff
@@ -0,0 +1,208 @@
...
+  lastCalledEndpoint: ?string;
+  previewsUrl: string;
+  transportMode: string;
+
+  accessToken = async (): Promise<AccessToken> =>
+    typeof this._optionAccessToken === "function"
+      ? this._optionAccessToken()
+      : this._optionAccessToken;
+
+  constructor(options: CommandOptions, client: Client) {
```
amccloud: These arguments seem to be flipped. I'd expect the required `client` to be first and the `options` to be second

jest.config.js
bitpshr: Yes, exactly: we mock a given CLI or API call, and then call the method in question. We then inspect the mocked CLI or API call and assert everything request-wise is correct: URLs, bodies, CLI arguments and order, etc. I didn't compare numerical coverage percentages (though I will create a ticket to do so), but we don't lose coverage from a theoretical point of view. It sounds bad, but we're basically doing manual snapshot testing.

package.json
bitpshr: Accident; added back in.

src/endpoints/BaseEndpoint.js
bitpshr: Interesting point; the client is actually only required if one method in one endpoint needs access to another method in another endpoint; it's sort of a hack which is why it was added second. I agree that these should be reversed though for clarity. Flipped.

src/testing.js
```diff
@@ -0,0 +1,63 @@
...
+  });
+}
+
+export function mockAPI(
+  url: string,
+  response: Object,
+  code: number = 200,
+  method: string = "get"
+) {
+  (nock("http://api"): any)[method](url).reply(code, response);
```
amccloud: What is the type without the cast? I would sorta expect it to already be any

src/endpoints/Comments.test.js
```diff
@@ -0,0 +1,44 @@
+// @flow
+import { mockAPI, API_CLIENT } from "../testing";
+
+describe("#create", () => {
+  test("api", async () => {
+    mockAPI("/projects/project/branches/branch", { name: "branch" });
+    mockAPI("/comments", { id: "1337" }, 201, "post");
```
amccloud: @bitpshr I see what you mean with the mocks. love it!

I think a few things that missing from the snapshot coverage would be:

- body for post request
- query params for get request
- headers

bitpshr:
I went through the remaining action items after yesterday's chat @tommoor and @amccloud:

## Double-proxy removal

Quick context: this is currently being used to cache the last-called method for informative errors. I looked into this and we have a few options:

1. We can get rid of the double proxy by getting the `callee` without using `arguments.callee` by doing the following:

    ```js
    const stack = new Error().stack;
    const callee = stack.split("\n")[1].replace(/(\(.*\)|at)/g, "").trim();
    ```

    Note that we'd have to do this in every method; we can't put this into a shared utility function because we'd lose the very stack trace information we're after. I personally think this is too ugly and hacky to be used in a module meant for production like our SDK.

2. We can explicitly pass the callee internally (note the last argument to `this.request`):

    ```js
    info(descriptor: ActivityDescriptor): Promise<Activity> {
        return this.request<Promise<Activity>>({
            api: () => {
                return this.apiRequest(`activities/${descriptor.activityId}`);
            }
        }, "info");
    }
    ```

    This feels burdensome since we're repeating a method name that's defined only a few lines above, all to overcome a `strict` mode limitation around `callee` for error reporting. Still, it's pretty minimal overhead.

3. We can let the double proxy handle this automatically and revisit this if debugging ever becomes an issue. This is my vote.

## Snapshot testing

I quickly got crude snapshot testing locally running using the same tests from `master` and comparing the snapshots. Things looked mostly-sane; nothing was unexpected or looked erroneous:
- [API snapshot diff](https://github.com/goabstract/abstract-sdk/commit/79a8ffcbf697d617070921c470db455db0c9aaa9#diff-69b565a34af9479f19f50762918685a0)
- [CLI snapshot diff](https://github.com/goabstract/abstract-sdk/commit/79a8ffcbf697d617070921c470db455db0c9aaa9#diff-0cf440d5529dfbfe55147eee70807b97)

## Breaking changes

1. Client creation becomes class instantiation.
2. The `transport` configuration option becomes `transportMode` and is a string, "cli" or "api".
3. `BranchDescriptor` no longer requires a `sha` property.
4. Some APIs that exected or returned `BranchDescriptor` now use `CommitDescriptor`.

## Prerelease

A prerelease is currently available in our private registry at `abstract-sdk@1.0.0-beta.3`.

## Next steps

Documentation has been updated (external changes were minimal) and CI will pass once we remove the temporary snapshot diffs. Beyond that, once @tommoor and @amccloud feel ready and confident, we can merge this and cut a `1.0.0` release.

.editorconfig
rickharris: Do we use prettier in this project?

.eslintignore
```diff
@@ -1,4 +0,0 @@
-node_modules
-flow-typed
-__snapshots__
-/lib
```
rickharris: was this unnecessary? does eslint use .gitignore?

.editorconfig
bitpshr: We do, but it's still jarring to start a new file and have it use my default editor settings. Prettier via linting fixes any issues, but doesn't prevent them from happening (e.g. hitting "TAB" will still insert a hard tab for me.) I don't lint on save like a lot of others, so this is more noticeable.

rickharris: > I don't lint on save like a lot of others

All makes sense now 🙃

.eslintignore
bitpshr: Yes, we now use `.gitignore`.

scripts/install-types.sh
```diff
@@ -0,0 +1,17 @@
...
+types=(
+  "jest@^23.x.x"
+  "query-string@^6.x.x"
+  "nock@^10.x.x"
+  "uuid@^3.x.x"
+)
+
+for i in "${types[@]}"
+do
+  yarn flow-typed install $i
```
amccloud: @bitpshr so I think this only works if you have a yarn and flow-typed installed in your project already. I guess when you tried `npm install` locally it was after `yarn install` setup flow-typed

src/Client.js
```diff
@@ -0,0 +1,94 @@
...
+    this.layers = new Layers(this, options);
+    this.notifications = new Notifications(this, options);
+    this.organizations = new Organizations(this, options);
+    this.pages = new Pages(this, options);
+    this.previews = new Previews(this, options);
+    this.projects = new Projects(this, options);
+    this.shares = new Shares(this, options);
+    this.users = new Users(this, options);
+
+    // This only for informative errors; we proxy each method
```
amccloud:
```suggestion
    // This is only for informative errors; we proxy each method
```

package.json
```diff
@@ -1,23 +1,27 @@
...
-  "version": "0.7.1",
+  "version": "1.0.0-beta.7",
   "description": "",
-  "files": [
-    "/lib",
-    "/docs"
-  ],
   "main": "lib/index.js",
   "browser": {
-    "./lib/AbstractCLI": false
```
amccloud: Can we move `@elasticprojects/abstract-cli` to `optionalDependencies` while we're at it to silence this warning

> npm WARN abstract-sdk@1.0.0-beta.7 requires a peer of @elasticprojects/abstract-cli@^1.0.0 but none is installed. You must install peer dependencies yourself.

src/types.js
```diff
@@ -1401,148 +1407,31 @@ export type CursorResponse<T> = {
...
   layers: Layer[]
 };
 
+export type CollectionMeta = {
+  collection: Collection,
+  files: File[],
+  pages: Page[],
+  layers: Layer[]
+};
+
```
amccloud: Maybe `CollectionResponse` and `CollectionsResponse` since it's pretty much the raw response? ...or `CollectionPage` to align with the page object model

src/index.js
```diff
@@ -1,27 +1,36 @@
...
   Error: BaseError,
+  FileAPIError,
   ForbiddenError,
   InternalServerError,
+  LatestCommitNotFound,
   NotFoundError,
   RateLimitError,
   ServiceUnavailableError,
   UnauthorizedError
 };
```
amccloud: wishlist: `export type * from "./errors"` since these all end in Error. Would be nice to write `Abstract.ForbiddenError` over `Abstract.errors.ForbiddenError`

src/index.js
```diff
@@ -1,27 +1,36 @@
...
 
-export { client, errors, Sketch, TRANSPORTS };
-export { client as Client }; // Deprecated: prefer Abstract.client factory
```
amccloud: 👌

src/Client.js
```diff
@@ -0,0 +1,94 @@
...
+  constructor(options: $Shape<CommandOptions> = {}) {
+    options = {
+      accessToken: process.env.ABSTRACT_TOKEN,
+      apiUrl: "https://api.goabstract.com",
+      cliPath: "node_modules/@elasticprojects/abstract-cli/bin/abstract-cli",
+      previewsUrl: "https://previews.goabstract.com",
+      transportMode: "api",
+      webUrl: "https://app.goabstract.com",
+      ...options
+    };
```
amccloud: 🏅

src/endpoints/Branches.test.js
```diff
@@ -0,0 +1,50 @@
+// @flow
+import { mockAPI, mockCLI, API_CLIENT, CLI_CLIENT } from "../testing";
+
+describe("#info", () => {
+  test("api", async () => {
+    mockAPI("/projects/project/branches/branch", { id: "1337" });
```
amccloud: I've been using `{ id: "branch-id" }` or `noun-id` respectively to make it easier to track some of these values while debugging and it's been really helpful

scripts/install-types.sh
bitpshr: This was an oversight on my part; we can just use `npx` and totally detach ourself from NPM's lifecycle.

src/index.js
bitpshr: Say no more; another oversight on my part. We were already explicitly importing the public errors we want to expose, so we could...just export them instead of assigning them to a `const`.

src/endpoints/Branches.test.js
bitpshr: My OCD won't let this comment sit; quest accepted  ⛰

bitpshr:
Thanks for the reviews so far @amccloud, really helpful. All feedback has been addressed. I cut a new `abstract-sdk@1.0.0-beta.8` pre-release and updated the embed PR to use it for verification at https://github.com/goabstract/ui/pull/569. Embeds still work as expected on that branch. We also got rid of `Abstract.errors.<some_error>`: https://github.com/goabstract/ui/pull/569/commits/2051550caef04118b2c26b8504b06c9cae72c115#diff-ccdfb9d333d9630a8d9b070ce7db7040R42. Great suggestion.

I also verified the `post-install` issue should be fixed by leveraging `npx` instead `yarn` or `npm`. Check the terminal session depicted below on a fresh clone of `abstract-graphql`:

screen shot 2019-03-01 at 4 38 44 pm

rickharris:
Are we purposely installing flow type defs when abstract-sdk is a dependency of the current project, or is that a side-effect of our flow-typed install approach? Will those actually do anything? I thought we only wanted them for working internally.

src/endpoints/Branches.test.js
bitpshr: All tests updated per your suggestion for easier debugging. This also points tot the need for unified descriptor generators like we had before. I will make a ticket to add that to the new structure.

bitpshr:
@rickharris: Good question. Theoretically, it's possible for an external SDK type to leverage (or even pass through) a 3rd-party type brought in by the SDK via `flow-typed`. Thinking about this more though, I doubt the Flow compiler or `flow-typed` even know to check `node_modules/<some_module>/flow-typed/` dirs when compiling a parent project unless `.flowconfig` is explicitly configured to do so.

We could probably change `postinstall` to `prepare`.

bitpshr:
I updated this to use `prepare` instead of `postinstall` and verified that this change only installs internal flow types during local installation. This is clean, and if we ever do expose an external API that relies on a `flow-type`, we can use `postinstall` specifically for public third-party typings. Good call @rickharris.

Embeds aren't heavily used yet but the latest development builds are relying on the beta SDK, so that's one testing ground for us. How do y'all feel about a potential path to land this work? I don't think there's any rush, but I do think we're in a bit of a holding pattern with major new development.

cc @amccloud  @tommoor

amccloud:
@bitpshr everything looks good on my end. Embed is ticking along just fine. It  looks I'll need to find another way to introspect the endpoint methods with the `abstract-graphql` project, but that seems doable and really out of scope to begin with

I wish the paginate stuff was a diffrent PR as I haven't had a chance to try that  out yet. Should we label it as experimental in the docs for now?

The only concern after that is communicating backward incompatible changes. Should probably add thew new API for `new Abstract.Client()` to the changelog. Anyting else we can add?

bitpshr:
Hi @amccloud. In hindsight, I definitely should've added in the asynchronous pagination function in a different PR, but I feel less-worried about it since it's literally [just an interface adaptation](https://github.com/goabstract/abstract-sdk/blob/d2e12c1881336b1f75e87e6995f9264db26379a4/src/paginate.js#L4-L20). I went ahead and marked it as experimental in the documentation as suggested, though we could also not document it at all yet if that's even safer. The CHAGELOG has been updated with the new client instantiation as well.

I think this should be ready for prime time whenever we all feel ready to push buttons!

bitpshr:
@amccloud: Thanks a ton for all the reviews. The last two commits since your approval (6ca6966 and a742a9d) added required babel dependencies and made `CollectionDescriptor.branchId` optional, respectively. I think this should be good to go still, but if you see anything amiss, don't hesitate to call it out.

Barring any newly-discovered issues, I plan to land this in the morning and to cut our first major release shortly thereafter.

cc @rickharris @tommoor

src/endpoints/Comments.js
```diff
@@ -0,0 +1,118 @@
...
+      }
+    });
+  }
+
+  list(
+    descriptor:
+      | BranchDescriptor
+      | CommitDescriptor
+      | LayerDescriptor
+      | PageDescriptor,
```
amccloud: So nice that this isn't a Shape of an intersection anymore 👍

src/endpoints/Comments.js
```diff
@@ -0,0 +1,118 @@
...
+    options: ListOptions = {}
+  ): CursorPromise<Comment[]> {
+    let newDescriptor;
+    return this.request<CursorPromise<Comment[]>>({
+      api: () => {
+        return new Cursor<Comment[]>(
+          async (meta = { nextOffset: options.offset }) => {
+            if (!newDescriptor) {
```
amccloud: Not sure I follow what is going on here with the condition. Could `newDescriptor` (maybe named `latestDescriptor`) be a const and defined above? It doesn't seem like it changes

bitpshr: This was interesting. Since this is a paginated `list` method, it's supposed to return `CursorPromise`. If we `await` or chain any additional `Promise`-returning methods, like `getLatestDescriptor`, we break the expected signature of this method. Lazily-checking the descriptor when fetching new pages allows this since no chaining has to happen before returning the `CursorPromise`.

src/types.js
```diff
@@ -26,38 +25,35 @@ export type NotificationDescriptor = {|
...
 
-type ObjectDescriptor = {|
+export type BaseCommitDescriptor = {
+  sha: string,
   projectId: string,
-  branchId: string | "master",
-  sha: string
-|};
+  branchId: string | "master"
+};
```
amccloud: Hard to see in this diff but this matches `CommitDescriptor` now

src/endpoints/Comments.js
```diff
@@ -0,0 +1,118 @@
...
+export default class Comments extends Endpoint {
+  async create(
+    descriptor:
+      | BranchDescriptor
+      | CommitDescriptor
+      | LayerDescriptor
+      | PageDescriptor,
+    comment: NewComment
+  ): Promise<Comment> {
+    if (descriptor.sha) {
```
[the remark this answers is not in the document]
bitpshr: In every other use of `getLatestDescriptor` so far, we're relying on the no-op; happy to change this up, but the no-op was nice to work with and felt clean.

src/endpoints/Commits.js
```diff
@@ -0,0 +1,81 @@
...
+    descriptor: T
+  ): Promise<T> {
+    if (descriptor.sha !== "latest") return descriptor;
+    const [commit] = await this.list((descriptor: any), { limit: 1 });
+    return {
+      ...descriptor,
+      commitSha: commit.sha,
+      sha: commit.sha
+    };
+  }
```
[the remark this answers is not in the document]
bitpshr: Good point. Moved to `client.descriptors`.

src/endpoints/Comments.js
```diff
@@ -0,0 +1,118 @@
...
+      api: async () => {
+        const branch = await this.client.branches.info({
+          branchId: descriptor.branchId,
+          projectId: descriptor.projectId
+        });
+        let commentData = {
+          branchName: branch.name
+        };
+        if (descriptor.layerId) {
+          const layer: any = await this.client.layers.info(descriptor);
```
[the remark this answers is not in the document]
bitpshr: Temporary insanity on my part. Good catch.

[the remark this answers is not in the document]
amccloud: @bitpshr I cleaned this up on my `pageId` branch

src/endpoints/Comments.js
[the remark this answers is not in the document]
amccloud: I like the no-op too 👍

Issue #101, opened 2019-03-07
unmet `abstract-cli` dependency


gerrit:
Running `yarn add abstract-sdk` (as per https://sdk.goabstract.com/docs/installation/) fails with this error message: `error Couldn't find package "@elasticprojects/abstract-cli@^1.0.0" required by "abstract-sdk@^1.0.0-beta.8" on the "npm" registry.`

manually pinning to 0.7.4 downgrades this to a warning (`warning " > abstract-sdk@0.7.4" has unmet peer dependency "@elasticprojects/abstract-cli@^1.0.0".)

Seems like the public `abstract-sdks` packages are depending on a private NPM package?

bitpshr:
Thanks for taking the time to file this @gerrit. Please note that all `abstract-sdk@1.0.0-beta.*` versions  are pre-releases of an internally-reorganized version of the SDK, with minor external breaking changes. We plan to communicate these changes to developers and to release updated documentation in the near future.

For now, we've updated the [feature/types-and-modules](https://github.com/goabstract/abstract-sdk/tree/feature/types-and-modules) branch of the SDK that powers these pre-releases, and cut a new version at `abstract-sdk@1.0.0-beta.9` that can be used for bleeding-edge testing. Still, it's currently recommended to use the `0.x.x` release line in production.

Thanks again for helping to preemptively identify a potentially serious issue.

gerrit:
Ah good to know. I just followed the public abstract documentation and ran `yarn add abstract-sdk` – so turns out that `yarn`, unlike `npm`, doesn't have a concept of pre-release versions, so you might want to add the latest stable version number to your instructions to be safe :)

PR #103, opened 2019-03-07
bugfix: move CLI dependency from peer to optional

bitpshr:
This pull request moves `@elasticprojects/abstract-cli` from `peerDependencies` to `optionalDependencies` so that `abstract-sdk@0.7.x` can be installed externally without error.

References https://github.com/goabstract/abstract-sdk/issues/101


bitpshr:
Not sure this will help: https://github.com/yarnpkg/yarn/issues/5366

Issue #76, opened 2019-01-16
Remove pageId from LayerDescriptor


amccloud:
Would be nice to remove `pageId` from `LayerDescriptor`

- [x] [Layer routes](https://github.com/goabstract/ui/pull/204/files#r246593467)
- [x] [Share links](https://github.com/goabstract/projects/pull/1507)
- [x] [comments.create](https://github.com/goabstract/abstract-sdk/pull/106)
- [x] [LayerDescriptor](https://github.com/goabstract/abstract-sdk/pull/106)

bitpshr:
This was resolved by @amccloud in #106.

PR #106, opened 2019-03-21
Remove `pageId` from LayerDescriptor

amccloud:
For https://github.com/goabstract/abstract-sdk/issues/76

- Revert descriptor types back to explicitly defined types. I originally avoided spread because I assumed the object descriptors wouldn't always be perfectly hierarchical like with the absence of pageId
- Added utility functions back to help explicitly convert objects into exact descriptors to avoid usage of `any`
- Added new utility for converting layer into `PageDescriptor` for commenting
- Updated `comments.create` so that when commenting on a layer `pageId` is loaded from the layer
- Remove `pageId` from `LayerDescriptor`

### DIFF
--- a/src/endpoints/Comments.js
+++ b/src/endpoints/Comments.js
@@ -1,5 +1,10 @@
 // @flow
 import querystring from "query-string";
+import {
+  objectBranchDescriptor,
+  objectFileDescriptor,
+  layerPageDescriptor
+} from "../utils";
 import Cursor from "../Cursor";
 import type {
   BranchDescriptor,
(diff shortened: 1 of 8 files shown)


amccloud:
@bitpshr it's the way to go to an exact object and avoid the cast through any. One of the primary goals of descriptors is to eliminate manually mapping params like this:

```js
const someNewParams = {
  someId: otherParams.someId,
  somethingUnrelatedButAddedBecauseItWasEasy: true
}
```

in favor of more explicit mapping like in the utilities:


```js
const someNewParams = otherParamsToSomeNewParams(otherParams);
```

PR #113, opened 2019-04-04
feature: do not preload entities when creating comments

### LINKED ISSUE #109: Remove preloading of names from comment create
The backend now supports loading `branchName`, `layerName` etc so they do not need to be passed to `comment.create` any more 👏

bitpshr:
This pull request updates the `comments.create` endpoint to no longer preload branches, layers, pages, or files before creating comments. This was previously required to gather information such as entity name, file type, etc., but this is no longer required after recent API updates.

Resolves #109

### DIFF
--- a/src/endpoints/Comments.js
+++ b/src/endpoints/Comments.js
@@ -1,10 +1,5 @@
 // @flow
 import querystring from "query-string";
-import {
-  objectBranchDescriptor,
-  objectFileDescriptor,
-  layerPageDescriptor
-} from "../utils";
 import Cursor from "../Cursor";
 import type {
   BranchDescriptor,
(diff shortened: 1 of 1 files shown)


src/endpoints/Comments.js
```diff
@@ -24,8 +19,11 @@ export default class Comments extends Endpoint {
     descriptor:
       | BranchDescriptor
       | CommitDescriptor
-      | LayerDescriptor
-      | PageDescriptor,
+      | PageDescriptor
+      | {|
+          ...$Exact<LayerDescriptor>,
+          pageId: string
+        |},
```
bitpshr: The API explicitly requires a `pageId` when a `layerId` is passed even though most other commands dealing with layer-specific actions only require a `layerId` and `fileId` (like a normal `LayerDescriptor`.) As such, we make sure we have a `pageId` when passing a `LayerDescriptor` here.

PR #114, opened 2019-04-04
feature: deploy documentation on branch

### LINKED ISSUE #62: Change documentation deployment to not happen on master
We often have work sat on master that hasn't yet been published to NPM as we're working towards a release. As it stands right now documentation updates related to this work get auto-deployed with master.

We've discussed:

- **Deploying docs when a release is tagged.** This one is great in that it requires no thinking but prevents us from updating docs outside of the release cycle.
- **Deploying docs by pushing to a branch.** This requires manual deployment but we can do so at any time.

Currently swaying towards the branch approach.

bitpshr:
This pull request updates the the continuous integration script so that documentation is only deployed on pushes to the `docs` branch. This will allow documentation updates independently of `master` pushes or any release (NPM, git, or otherwise.) Note that this requires one extra step when cutting a new release: rebasing `master` on top of `docs` to pick up any typo fixes or changes made directly against the `docs` branch.

Resolves #62

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -102,7 +102,7 @@ workflows:
             - test
           filters:
             branches:
-              only: master
+              only: docs
       - publish:
           <<: *tagged
           requires:
--- a/README.md
+++ b/README.md
@@ -37,6 +37,7 @@ Changes and additions to the SDK should include tests and documentation updates 
 ## Publishing
 
 ```bash
+$ git rebase docs
 $ yarn version
 $ git push --follow-tags
 $ npm publish


tommoor:
I was actually kind of into the docs being deployed from master. Why do we care about deploying docs without pushing to master – is there some situation you've come across that prompted the PR?

bitpshr:
@tommoor: I am a little confused; this PR was based on the issue you filed at https://github.com/goabstract/abstract-sdk/issues/62. The documentation site should reflect the latest stable release, not the latest work on `master`. How would we push a documentation hotfix without also deploying unreleased documentation that may be on `master`?

tommoor:
oh no

PR #115, opened 2019-04-05
bugfix: check peerDependencies and MacOS app folder for CLI

bitpshr:
This pull request updates the `Client` to check for a CLI executable in both `peerDependencies` and the macOS Abstract application folder.  This functionality existed before the reorganization in #98 and was lost in translation.

### DIFF
--- a/src/Client.js
+++ b/src/Client.js
@@ -1,4 +1,5 @@
 // @flow
+import locatePath from "locate-path";
 import Activities from "./endpoints/Activities";
 import Assets from "./endpoints/Assets";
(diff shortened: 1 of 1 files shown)


src/Client.js
```diff
@@ -1,4 +1,5 @@
 // @flow
+import locatePath from "locate-path";
```
bitpshr: We already had `locate-path` in the `package.json` dependency list, so we explicitly brought it over during the reorganization. Total oversight on my part not to use it.

PR #116, opened 2019-04-05
feature: implement decorator-based entity cache

### LINKED ISSUE #87: Implement a request cache
We should cache requests where possible to improve perceived performance and to reduce load on our servers.

bitpshr:
This pull request implements an entity cache for static response values, or response values that will never change over time. This is especially useful for methods that always return the same entity based on a given descriptor (like `info` methods), whereas methods that inherently can't be cached since they must return up-to-date, dynamic data reap no benefit (like `list` methods).

The cache was implemented as a decorator and makes use of aspect-oriented programming to wrap normal requests at runtime.

Resolves #87

### DIFF
--- a/src/Client.js
+++ b/src/Client.js
@@ -42,12 +42,15 @@ export default class Client {
   shares: Shares;
   users: Users;
 
+  cache = new Map<string, any>();
+
   constructor(options: $Shape<CommandOptions> = {}) {
     options = {
       accessToken: process.env.ABSTRACT_TOKEN,
       apiUrl: "https://api.goabstract.com",
       cliPath:
         "/Applications/Abstract.app/Contents/Resources/app.asar.unpacked/node_modules/@elasticprojects/abstract-cli/bin/abstract-cli",
+      maxCacheSize: 0,
       previewsUrl: "https://previews.goabstract.com",
       transportMode: "api",
       webUrl: "https://app.goabstract.com",
--- a/src/endpoints/Activities.js
+++ b/src/endpoints/Activities.js
@@ -17,6 +17,10 @@ export default class Activities extends Endpoint {
     return this.request<Promise<Activity>>({
       api: () => {
         return this.apiRequest(`activities/${descriptor.activityId}`);
+      },
+
+      cache: {
+        key: `activity:${descriptor.activityId}`
       }
     });
   }
--- a/src/endpoints/Assets.js
+++ b/src/endpoints/Assets.js
@@ -10,17 +10,23 @@ export default class Assets extends Endpoint {
         return this.apiRequest(
           `projects/${descriptor.projectId}/assets/${descriptor.assetId}`
         );
+      },
+
+      cache: {
+        key: `asset:${descriptor.assetId}`
       }
     });
   }
 
   async list(descriptor: CommitDescriptor): Promise<Asset[]> {
-    descriptor = await this.client.descriptors.getLatestDescriptor(descriptor);
+    const latestDescriptor = await this.client.descriptors.getLatestDescriptor(
+      descriptor
+    );
     return this.request<Promise<Asset[]>>({
       api: async () => {
-        const query = querystring.stringify({ sha: descriptor.sha });
+        const query = querystring.stringify({ sha: latestDescriptor.sha });
         const response = await this.apiRequest(
-          `projects/${descriptor.projectId}/assets?${query}`
+          `projects/${latestDescriptor.projectId}/assets?${query}`
         );
         return response.data.assets;
       }
@@ -42,6 +48,10 @@ export default class Assets extends Endpoint {
           },
           null
         );
+      },
+
+      cache: {
+        key: `asset-raw:${descriptor.assetId}`
       }
     });
   }
--- a/src/endpoints/Changesets.js
+++ b/src/endpoints/Changesets.js
@@ -4,13 +4,15 @@ import Endpoint from "./Endpoint";
 
 export default class Changesets extends Endpoint {
   async info(descriptor: CommitDescriptor): Promise<Changeset> {
-    descriptor = await this.client.descriptors.getLatestDescriptor(descriptor);
+    const latestDescriptor = await this.client.descriptors.getLatestDescriptor(
+      descriptor
+    );
     return this.request<Promise<Changeset>>({
       api: async () => {
         const response = await this.apiRequest(
-          `projects/${descriptor.projectId}/branches/${
-            descriptor.branchId
-          }/commits/${descriptor.sha}/changeset`
+          `projects/${latestDescriptor.projectId}/branches/${
+            latestDescriptor.branchId
+          }/commits/${latestDescriptor.sha}/changeset`
         );
         return response.changeset;
       },
@@ -18,12 +20,17 @@ export default class Changesets extends Endpoint {
       cli: () => {
         return this.cliRequest([
           "changeset",
-          descriptor.projectId,
+          latestDescriptor.projectId,
           "--commit",
-          descriptor.sha,
+          latestDescriptor.sha,
           "--branch",
-          descriptor.branchId
+          latestDescriptor.branchId
         ]);
+      },
+
+      cache: {
+        key: `changeset:${descriptor.sha}`,
+        disabled: descriptor.sha === "latest"
       }
     });
   }
--- a/src/endpoints/Endpoint.js
+++ b/src/endpoints/Endpoint.js
@@ -30,7 +30,7 @@ export type CacheConfiguration = {
   key: string
 };
 
-export type EndpointHandler<T> = {
+export type EndpointRequest<T> = {
   api?: () => T,
   cli?: () => T,
   cache?: CacheConfiguration
@@ -63,29 +63,30 @@ export default class Endpoint {
     this.webUrl = options.webUrl;
   }
 
-  request<T>(handler: EndpointHandler<T>): T {
+  request<T>(request: EndpointRequest<T>): T {
     let response;
 
-    if (handler.cache) {
-      const existingEntity = this.client.cache.get(handler.cache.key);
+    if (request.cache) {
+      const existingEntity = this.client.cache.get(request.cache.key);
       if (existingEntity) {
         return existingEntity;
       }
     }
 
     if (this.transportMode === "auto") {
-      if (handler.cli) {
-        response = handler.cli();
-      } else if (handler.api) {
-        response = handler.api();
+      if (request.cli) {
+        response = request.cli();
+      } else if (request.api) {
+        response = request.api();
       } else {
         throw new EndpointUndefinedError(
           this.lastCalledEndpoint,
           this.transportMode
         );
       }
-    } else if (handler[this.transportMode]) {
-      response = handler[this.transportMode]();
+    } else if (request[this.transportMode]) {
+      const handler = request[this.transportMode];
+      response = handler();
     } else {
       throw new EndpointUndefinedError(
         this.lastCalledEndpoint,
@@ -93,8 +94,8 @@ export default class Endpoint {
       );
     }
 
-    if (handler.cache && this.maxCacheSize > 0 && !handler.cache.disable) {
-      this.client.cache.set(handler.cache.key, response);
+    if (request.cache && this.maxCacheSize > 0 && !request.cache.disable) {
+      this.client.cache.set(request.cache.key, response);
 
       if (this.client.cache.size > this.maxCacheSize) {
         const oldestEntity = this.client.cache.keys().next().value;
--- a/src/endpoints/Endpoint.js
+++ b/src/endpoints/Endpoint.js
@@ -25,9 +25,15 @@ const logCLIRequest = log.extend("AbstractCLI:request");
 const logCLIResponse = log.extend("AbstractCLI:response");
 const minorVersion = version.split(".", 2).join(".");
 
+export type CacheConfiguration = {
+  disable?: boolean,
+  key: string
+};
+
 export type EndpointHandler<T> = {
   api?: () => T,
-  cli?: () => T
+  cli?: () => T,
+  cache?: CacheConfiguration
 };
 
 export default class Endpoint {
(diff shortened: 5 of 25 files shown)


src/endpoints/Activities.js
```diff
@@ -13,6 +14,10 @@ import type {
 import Endpoint from "./Endpoint";
 
 export default class Activities extends Endpoint {
+  @cache({
+    namespace: "activities",
+    requestId: "activityId"
+  })
```
bitpshr: This is an example of using the decorator. In most cases, only a `requestId` argument needs to be passed that specifies which property on the request descriptor should be used to uniquely identify this entity.

src/endpoints/Activities.js
```diff
@@ -13,6 +14,7 @@ import type {
 import Endpoint from "./Endpoint";
 
 export default class Activities extends Endpoint {
+  @cache("activityId")
```
bitpshr: This is an example of using the decorator.  The first (and in most cases, only) argument specifies which property on the descriptor uniquely identifies this entity.

tommoor:
Smart stuff. It's difficult for me to have enough confidence to approve however in the absence of tests – perhaps we can spy on the network request and ensure that duplicate request only request network once?

src/endpoints/Activities.js
amccloud: I think in most cases we'll want to cache based on the primary id and sha. I think we'll also need to account for the scenario where an endpoint takes in various types of arguments for the different descriptor inputs

To avoid the unkknown complexity that might arise from flow and other tooling not supporting decorators, what do you think about passing the endpoint's descriptor argument - the identity of the request - to `this.request` and let it handle the caching?

I think we can inconsequently include the other non-primary ids into the cache key generation. With that I think each endpoint can generate a valid cache key with `Object.values(arguments[0]).join(':')` (assuming the first argument is a descriptor)

Also, if not `this.request` what about our `Client` proxy since the argument position is so well known

bitpshr: I agree with pretty much everything you laid out above, especially around cache ID generation. I was struggling to think of the best way to uniquely identify an asset, and you're right: a stringified descriptor is the identity of the request, so we should use that for the cache ID. We still need to include the name of the endpoint as an added differentiator though since different endpoints can accept identical descriptors and return different objects, but the descriptor bit is what I was looking for. Thanks for this.

Candidly, I'm a huge fan of decorators. Angular, TypeScript, and other major tools have used or supported them for years, and they just underwent a recent (and potentially final) round of stage 2 proposal feedback. This decorator is also only used internally, so we control our decorator transpilation and users never know about it. Still, I'm happy to change the implementation so `Endpoint` exposes a new `cachedRequest` method alongside `request` that caches values in a similar manner.

I'll rework this now to see how that feels.

amccloud: Glad to see we're on the same page. While we're still trying to break ui out into sdk I think we should keep the projects as similar as possible that way we minimize our risk porting code between them

I've used decorators in other languages so I'm a fan when I know an API is rather concrete. But I've also experienced some degree of friction when refactoring code that uses decorators so I usually expose decorators as a public API instead of using on the internals

Another option to `cachedRequest` could be to make `request` smart and enable caching when it detects `sha` in the descriptor. There might be some complexity with `sha: "latest"`. That with an additional option in the second argument to completly control/opt-out of caching should really go far 🚀

src/Cache.js
```diff
@@ -0,0 +1,73 @@
...
+  cache = new Map<string, any>();
+  maxCacheSize: number;
+
+  constructor(client: Client, maxCacheSize: number) {
+    this.maxCacheSize = maxCacheSize;
+    return this._createCache(client);
+  }
+
+  _createCache(client: Client) {
+    return new Proxy(client, {
```
bitpshr: The endpoint proxy was pulled out of `Client` and used as the basis for this new class.

src/Cache.js
```diff
@@ -0,0 +1,73 @@
...
+        if (typeof client[endpointName] === "object" && client[endpointName]) {
+          return new Proxy(client[endpointName], {
+            get: (endpoint: Object, methodName: string) => {
+              if (
+                typeof endpoint[methodName] === "function" &&
+                !Endpoint.prototype.hasOwnProperty(methodName)
+              ) {
+                const method = endpoint[methodName];
+                endpoint.lastCalledEndpoint = `${endpointName}.${methodName}`;
+                if (method.isCached) {
```
bitpshr: This class does what the decorator did in the previous implementation: it maintains a `Map` and checks for existing entities before initiating underlying requests.

src/Cache.js
```diff
@@ -0,0 +1,73 @@
...
+                  return async (...args) => {
+                    if (args[0] && args[0].sha === "latest") {
+                      return method.apply(endpoint, args);
+                    }
+                    const cacheId = this._createCacheId(
```
bitpshr: Cache IDs are now generated using the following logic:

```js
const descriptorValues = Object.values(descriptor).join(":");
return `${endpointName}:${methodName}:${descriptorValues}`;
```

src/endpoints/Activities.js
```diff
@@ -10,16 +10,19 @@ import type {
...
 
 export default class Activities extends Endpoint {
-  info(descriptor: ActivityDescriptor): Promise<Activity> {
-    return this.request<Promise<Activity>>({
-      api: () => {
-        return this.apiRequest(`activities/${descriptor.activityId}`);
-      }
-    });
-  }
+  info = cached(
```
bitpshr: We need some way to represent that a method should or should not be cached. Looking for `sha` in a descriptor is not sufficient since many cacheable requests do not contain a `sha` (like `activities.info`). On a stylistic level, I also think some level of explicitness is good when enabling caching, rather than all functions magically being cached based on their descriptor, which feels a little magical.

The `cached` function takes a function reference, decorates it with a boolean to indicate the function reference is cacheable, then returns the reference.

bitpshr:
Thanks for the review so far @amccloud and @tommoor. This should be ready for another pass. I ended up going with a proxy-based approach as suggested instead of meddling with `Endpoint#request` or adding `Endpoint#cachedRequest`. The `Endpoint` class intentionally has no insight into descriptors or other endpoint-specific implementation details, and the proxy-based approach allowed us to maintain this encapsulation.

When all was said and done, this approach felt very similar to the decorator approach, with the added bonus of not being subject to upstream specification churn (good call @amccloud.) We sort of got lucky that we already had the exact style of double-proxy we needed to make this work, otherwise it would've felt heavy.

bitpshr:
@amccloud: This is ready for another review based on the approach we discussed. Thanks again for helping to fine-tune this API; this implementation feels really nice to work with.

bitpshr:
Any additional comments on this @amccloud @tommoor? I know it's a large one; most of the changes are just adding the use of the actual cache on different endpoints. Happy to chat through it if that's easier too.

src/Client.js
```diff
@@ -51,6 +53,7 @@ export default class Client {
         "node_modules/@elasticprojects/abstract-cli/bin/abstract-cli",
         "/Applications/Abstract.app/Contents/Resources/app.asar.unpacked/node_modules/@elasticprojects/abstract-cli/bin/abstract-cli"
       ]),
+      maxCacheSize: 1337,
```
amccloud: I realize as I read this I'm not sure what units this is in and can't find where it is used

src/Client.js
```diff
@@ -43,6 +43,8 @@ export default class Client {
   shares: Shares;
   users: Users;
 
+  cache = new Map<string, any>();
```
amccloud: Using `mixed` here could give you a shot at reading values out of the cache without casting

src/endpoints/Assets.js
```diff
@@ -42,6 +48,10 @@ export default class Assets extends Endpoint {
           },
           null
         );
+      },
+
+      cache: {
+        key: `asset-raw:${descriptor.assetId}`
```
amccloud: Oh this is a good one. Since we include the method name in `ui` request ids, maybe we should always include it on all of the cache keys for consistency?

Maybe `raw:asset:${descriptor.assetId}` or at this point `assets.raw:${descriptor.assetId}`?

src/endpoints/Changesets.js
```diff
@@ -4,26 +4,33 @@ import Endpoint from "./Endpoint";
...
+          latestDescriptor.sha,
           "--branch",
-          descriptor.branchId
+          latestDescriptor.branchId
         ]);
+      },
+
+      cache: {
+        key: `changeset:${descriptor.sha}`,
+        disabled: descriptor.sha === "latest"
```
amccloud: ❤️

src/endpoints/Shares.js
```diff
@@ -28,9 +28,20 @@ export default class Activities extends Endpoint {
   }
 
   info<T: Share>(descriptor: ShareDescriptor): Promise<T> {
+    let cacheKey = "";
+    if (descriptor.url) {
+      cacheKey = `share:${descriptor.url}`;
```
amccloud: Can use shareId here too with `share:${inferShareId(descriptor)}`

src/endpoints/Assets.js
bitpshr: I went with including the asset type instead of the method name since we're shooting for a pure entity cache, not a request or method cache. It's possible that a future method could also cause a raw asset to be cached by side effect, which should theoretically be the same raw assets cached here. If we differentiated by method name, they'd be different.

src/endpoints/Shares.js
bitpshr: Ah, very nice.

src/Client.js
bitpshr: Ah, forgot to add this logic in. Updated.

bitpshr: Good catch; I added this missing logic to respect `maxCacheSize` (which is just a maximum size on the underlying `Map`, so there are no units.) I will be sure to document this option in the new documentation updates I'm making at #124, which overhauls the configuration page.

src/Client.js
bitpshr: I had issues with `mixed` in this case. Since we return cached items at https://github.com/goabstract/abstract-sdk/pull/116/files#diff-d911039e0fdb24bcc0da7b1aa365b2aeR70, we'd have to somehow refine these cached values down to `T`, which is the return value of `Endpoint#request`. I wasn't sure how to actually refine a `mixed` type to a generic with unknown properties like `T`. Candidly, I haven't used `mixed` that much, so I could be missing something simple here, but it seemed like we had to use `any` to still allow generic use of `request`.

bitpshr:
Thanks again @amccloud! I owe you several large reviews. All feedback has either been addressed or responded to.  Good call re: using `inferShareID` when creating `Share` cache keys.

src/Client.js
amccloud: `any` is considered a type blackhole in flow. For downstream use it shouldn't be possible to type `any` as anything else but `any`

I think by assigning the generic to an `any` value, `T` just becomes `any`. To refine the placeholder you can type the placeholder `T: mixed`. Once you have `T` constrained you should be able to refine if necessary

src/Client.js
amccloud: @bitpshr @tommoor should we ship this with the cache turned off by default for a while? Would allow people to migrate/opt-in as necessary

src/Client.js
bitpshr: I'm not sure I follow. Specifically, we would need to somehow tell Flow that items coming out of the cache [here](https://github.com/goabstract/abstract-sdk/pull/116/files#diff-d911039e0fdb24bcc0da7b1aa365b2aeR70) are actually type `T`. How would we refine this `mixed` type to type `T`, even if we constrain the generic to `T: mixed`? I tried this and still got an error, and wasn't sure how to "refine" to T since it can be a number of different entity types. There are no properties to check for on T since it's generic. Do you have an example of what you mean?

Also, what's the actual issue in this case? The only thing that loses type safety here is the cache, and that's intentional; items can be a number of different entity types. To make sure we use it safely, we tell Flow that the items we retrieve from it are definitely type `T`, which we know for certain since we put them in the cache ourselves, in this very function. This `any` is never seen downstream since we tell Flow it's a T before any value leaves `Endpoint#request`.

What type safety is lost by using `any` in this case?

bitpshr: Ping on this comment @amccloud - I think the `mixed` vs. `any` Map is the only thing remaining before we can land this. I'm fine with changing the map `mixed`, but how do I then retrieve the item from the map and return it as `T` from `Endpoint#request`?

I'm still not sure why the `any` Map is an issue: we only use values from it in `Endpoint#request`, and Flow is told that those values are `T` before returning. I'm struggling to see where this isn't type safe to any consumer of `Endpoint#request`.

src/testing.js
```diff
@@ -33,6 +33,23 @@ export const CLI_CLIENT = new Client({
...
+  maxCacheSize: 1337,
+  previewsUrl: "http://previews",
+  transportMode: "api"
+});
+
+export const CLI_CLIENT_CACHED = new Client({
+  accessToken: "token",
+  apiUrl: "http://api",
+  cliPath: ".",
+  maxCacheSize: 1337,
```
amccloud: 🧶 Can properties be omitted if it's the same as the default? Or something low like `1`, so that it's a bit more assertable in test?

Also would be nice to continue to use the key name as stub value for strings: `accessToken: "accessToken"`, etc.

src/endpoints/Endpoint.js
```diff
@@ -50,37 +57,52 @@ export default class Endpoint {
...
+      } else if (handler.api) {
+        response = handler.api();
+      } else {
+        throw new EndpointUndefinedError(
+          this.lastCalledEndpoint,
+          this.transportMode
+        );
+      }
+    } else if (handler[this.transportMode]) {
+      response = handler[this.transportMode]();
```
amccloud: 🧶 `handler` -> `request`. What initially caught my eye was that I could technically set transport mode to `cache` and sorta get this to work. That part doesn't really matter but now that the handler is more like a description of our request semantics it sorta seems like it should be renamed:

```js
const handler = request[this.transportMode]; // Maybe one day validate `transportMode` when the client initiates
const response = handler();
```

PR #117, opened 2019-04-09
bugfix: update asset raw file documentation

bitpshr:
This pull request fixes the documentation for retrieving an assets file by adding a missing `await` keyword to the example code.


bitpshr:
Because an end-user noticed this issue and because the fix is a one-word change to example code, I'm going to make an exception and land this without review.

Issue #108, opened 2019-03-27
Access "role" via the user object


naughtandcross:
I'd love the ability to filter data by user role, i.e Admin vs Contributor vs Viewer. I can make some educated guess by analyzing each users Activity by type, but since we are already able to sort by these fields in app via the People tab, I figured I would inquire about the feasibility of getting this added to the API first. 😎

tommoor:
Hey @naughtandcross – Abstract users are global so the role depends on the organization that is being queried, we do have the concept of memberships that includes this information though.

@bitpshr it WOULD be nice if we were able to provide a way to access the role without having to manually join memberships data on the clientside. Any thoughts on what an interface for that could look like?

bitpshr:
Hi @naughtandcross, thanks for taking the time to file this feature request. As of `abstract-sdk@1.2.0`, you can now access [memberships](https://sdk.goabstract.com/docs/abstract-api/#memberships) that are specific to a project or an organization. These membership objects contain a `role`, the `User` that the membership represents, and other useful information.

Please review this new functionality and don't hesitate to let us know if you any issues, questions, or feedback.

PR #120, opened 2019-04-12
feature: add memberships endpoint to the API transport

bitpshr:
This pull request adds a new `Memberships` endpoint with API transport support.

Requires https://github.com/goabstract/projects/pull/1977
References #108

### DIFF
--- a/docs/abstract-api.md
+++ b/docs/abstract-api.md
@@ -722,7 +722,7 @@ abstract.memberships.list({
 
 ### Retrieve a membership
 
-`memberships.info(MembershipDescriptor): Promise<Membership>`
+`memberships.info(OrganizationMembershipDescriptor | ProjectMembershipDescriptor): Promise<Membership>`
 
 Load the info for a specific member
 
@@ -1235,12 +1235,20 @@ Reference for the parameters required to load resources with Abstract SDK.
 }
 ```
 
-### MembershipDescriptor
+### OrganizationProjectDescriptor
+
+```js
+{
+  projectId?: string,
+  userId: string
+}
+```
+
+### ProjectMembershipDescriptor
 
 ```js
 {
   organizationId?: string,
-  projectId?: string,
   userId: string
 }
 ```
--- a/docs/abstract-api.md
+++ b/docs/abstract-api.md
@@ -691,6 +691,49 @@ abstract.layers.info({
 ```
 
 
+## Memberships
+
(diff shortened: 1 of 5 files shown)


docs/abstract-api.md
```diff
@@ -1191,3 +1234,13 @@ Reference for the parameters required to load resources with Abstract SDK.
   projectId: string
 }
 ```
+
+### MembershipDescriptor
+
+```js
+{
+  organizationId?: string,
+  projectId?: string,
```
amccloud: I think we can drop these optional props and rename these to `UserDescriptor`. Then the endpoints can be typed as: 

```js
list(objectDescriptor: OrganizationDescriptor | ProjectDescriptor | UserDescriptor) {}

// and

info(userDescriptor: UserDescriptor) {}
```

bitpshr: Much cleaner, great suggestion. Updated.

docs/abstract-api.md
```diff
@@ -1191,3 +1234,21 @@ Reference for the parameters required to load resources with Abstract SDK.
...
+  userId: string
+}
+```
+
+### ProjectMembershipDescriptor
+
+```js
+{
+  organizationId?: string,
```
amccloud: Does `organizationId` and `projectId` (above) still need to be nullable?

src/types.js
```diff
@@ -328,6 +328,27 @@ export type User = {
...
+
+export type ProjectMembershipDescriptor = {|
+  ...$Exact<MembershipDescriptorBase>,
+  projectId: string
+|};
+
+export type OrganizationMembershipDescriptor = {|
+  ...$Exact<MembershipDescriptorBase>,
+  organizationId: string
+|};
```
amccloud: Do we still need `MembershipDescriptorBase`? Seem like `userId` can just be hardcoded in each type

src/types.js
```diff
@@ -328,6 +328,27 @@ export type User = {
...
+
+export type Membership = {
+  createdAt: string,
+  organizationId?: string,
+  projectId?: string,
```
amccloud: When are these undefined? Seems like at least `organizaionId` should always be set

bitpshr: Good catch. `projectId` can be undefined for an organization-only membership, but you're totally right about `organizationId` always being populated. Updated.

PR #126, opened 2019-04-24
feature: add polling-based CI documentation

bitpshr:
This pull request adds documentation on polling-based continuous integration and an example Slack integration.

preview


bitpshr:
@tommoor: I totally agree that "continuous integration" is an odd title. I see your point about "web hooks" being a more search-friendly title, but I worried consumers would arrive on the page and read it, only to be let down / frustrated. I figured we'd change this page title as soon as web hooks support does actually land.

PR #129, opened 2019-05-03
bugfix: use lastChangedAtSha for latest preview requests

bitpshr:
This pull request brings the SDK up to parity with the desktop and web applications with how they render layer preview images. Specifically, the applications default to `layer.lastChangedAtSha`, whereas the SDK was defaulting to the latest layer commit's `sha`. We noticed a timing issue when using the SDK's approach; some embeds would appear broken because the latest commit's `sha` didn't yet have a preview image generated. Later in the day, with no changes, they started to work.

For now, so the SDK matches the application exactly, we're using `layer.lastChangedAtSha` for latest preview requests.

Fixes https://goabstract.atlassian.net/browse/PLATFORM-273

### DIFF
--- a/src/endpoints/Descriptors.js
+++ b/src/endpoints/Descriptors.js
@@ -5,13 +5,32 @@ import Endpoint from "./Endpoint";
 export default class Descriptors extends Endpoint {
   async getLatestDescriptor<T: ObjectDescriptor>(descriptor: T): Promise<T> {
     if (descriptor.sha !== "latest") return descriptor;
+
     const [commit] = await this.client.commits.list((descriptor: any), {
       limit: 1
     });
-    return {
+
+    let latestDescriptor: any = {
       ...descriptor,
       commitSha: commit.sha,
       sha: commit.sha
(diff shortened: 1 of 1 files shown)


bitpshr:
Thanks for the offline feedback @amccloud, this should be ready for a proper review now. Specifically, this PR updates `Descriptors#getLatestDescriptor` to special-case `LayerDescriptors` as discussed. Now, any time a latest `LayerDescriptor` is resolved, regardless of which endpoint triggers the resolution, the returned descriptor will use `lastChangedAtSha` instead of `sha`.

This is an immediate fix for an active bug plaguing embeds and keeps all resolutions of latest `LayerDescriptors` the same, rather than only special-casing previews. I'm still digging into why the underlying commits API doesn't return what we expect, but the CLI works the same in this regard.

bitpshr:
After talking through this more with @amccloud, this PR should be unnecessary based on the way commits are returned for `LayerDescriptor` objects from both the API and the CLI.

PR #130, opened 2019-05-03
feature: add files.raw to the CLI transport

bitpshr:
This pull request does the following:
- Modifies all `raw` methods to save to disk by default (per @tommoor)
- Adds `RawOptions` to allow opting-out of disk saving and specifying a custom filename
- Adds `files.raw` support to the CLI transport
- Cleans up duplicated of return types in each endpoint
- Updates documentation so API and CLI indicators are per-method, not per-endpoint

Resolves https://goabstract.atlassian.net/browse/PLATFORM-248

### DIFF
--- a/docs/abstract-api.md
+++ b/docs/abstract-api.md
@@ -9,8 +9,6 @@ title: Client
 
 ## Activities
 
-![API][api-icon]
-
 An activity represents a designated type of event within a project. These events can be specific to the project itself, or they can be specific to a collection, a branch, a commit, or a review within the project.
 
 ### The activity object
@@ -26,6 +24,8 @@ An activity represents a designated type of event within a project. These events
 
 ### List all activities
 
+![API][api-icon]
+
 `activities.list(BranchDescriptor | OrganizationDescriptor | ProjectDescriptor, { limit?: number, offset?: number }): Promise<Activity[]>`
 
 List the first two activities for a given project on a specific branch
@@ -38,6 +38,8 @@ abstract.activities.list({
 ```
 
 ### Retrieve an activity
+
+![API][api-icon]
 
 `activities.info(ActivityDescriptor): Promise<Activity>`
 
@@ -55,8 +57,6 @@ abstract.activities.info({
 
 
 ## Assets
-
-![API][api-icon]
 
 An asset represents a resource exported from a design file. Assets are automatically updated and available for new commits.
 
@@ -83,6 +83,8 @@ An asset represents a resource exported from a design file. Assets are automatic
 
 ### List all assets
 
+![API][api-icon]
+
 `assets.list(BranchDescriptor): Promise<Asset[]>`
 
 List all assets for a commit on a given branch of a project
@@ -97,6 +99,8 @@ abstract.assets.list({
 
 ### Retrieve an asset
 
+![API][api-icon]
+
 `asset.info(AssetDescriptor): Promise<Asset>`
 
 Load the info for an asset
@@ -109,6 +113,8 @@ abstract.assets.info({
 ```
 
 ### Retrieve an asset file
+
+![API][api-icon]
 
 `asset.raw(AssetDescriptor, { filename?: string, disableWrite?: boolean }): Promise<ArrayBuffer>`
 
@@ -142,8 +148,6 @@ fs.writeFile("asset.png", Buffer.from(processedBuffer), (err) => {
 
 ## Branches
 
-![CLI][cli-icon] ![API][api-icon]
-
 A branch is where design work and commits happen. A branch acts as a personal workspace for contributors, we encourage branches
 to be created for logical chunks of work – for example designing a new feature.
 
@@ -169,6 +173,8 @@ to be created for logical chunks of work – for example designing a new feature
 
 ### List all branches
 
+![CLI][cli-icon] ![API][api-icon]
+
 `branches.list(ProjectDescriptor, { filter?: "active" | "archived" | "mine" }): Promise<Branch[]>`
 
 List the active branches for a project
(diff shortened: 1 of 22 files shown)


src/endpoints/Activities.js
```diff
@@ -13,7 +13,7 @@ import type {
 import Endpoint from "./Endpoint";
 
 export default class Activities extends Endpoint {
-  info(descriptor: ActivityDescriptor): Promise<Activity> {
+  info(descriptor: ActivityDescriptor) {
```
bitpshr: These changes are throughout this PR; we were explicitly specifying a return type when Flow already deduced this correctly for every method we have so far.

src/endpoints/Assets.js
```diff
@@ -48,6 +54,14 @@ export default class Assets extends Endpoint {
...
           null
         );
+
+        if (!options.disableWrite) {
+          const filename =
+            options.filename || `${asset.layerName}.${asset.fileFormat}`;
+          fs.writeFile(filename, Buffer.from(arrayBuffer));
+        }
+
+        return arrayBuffer;
```
bitpshr: Even though we're writing to disk by default (unless `options.disableWrite`) is passed, it makes sense to always return the `arrayBuffer` so the underlying raw value can be always be cached.

src/endpoints/Files.js
```diff
@@ -60,4 +65,26 @@ export default class Files extends Endpoint {
       }
     });
   }
+
+  async raw(descriptor: FileDescriptor, options: RawOptions = {}) {
```
bitpshr: This is the main addition in this PR.

docs/abstract-api.md
```diff
@@ -26,6 +24,8 @@ An activity represents a designated type of event within a project. These events
 
 ### List all activities
 
+![API][api-icon]
```
bitpshr: All indicators were moved to be per-method rather than per-endpoint since we now have endpoints with methods that support different transports.

tommoor:
What do you think about combining the `fileName` and `disableWrite` option, so that not passing a `fileName` parameter gives you the data back in a buffer rather than writing the file to the current directory.

bitpshr:
I thought about that initially @tommoor but felt it would be too restrictive since it forces a consumer to specify a filename instead of letting them use whatever the name (or ID) of the retrieved asset is. See the comment at https://github.com/goabstract/abstract-sdk/pull/130#discussion_r280883573 too re: always returning data (if possible) to assist with caching.

tommoor:
Could also provide `filePath` and `fileName` separately if that's a concern. Something not quite sitting well about the disableWrite option – would love another opinion. cc @amccloud

bitpshr:
FWIW, Node's filesystem APIs do not separate `filename` and `path` - you supply a single string value that is both the path and filename. I was trying to stay 1:1 with Node APIs since so many consumers are used to them already.

I'm fine with changing the API to check for `filePath` and `fileName` and if either are present, to save to disk. I'm a little confused what this buys us though - we still have two options, and they ultimately determine if files are written to disk. What's wrong with `disableWrite`? The name?

bitpshr:
Ping on this one @tommoor and @amccloud. I personally like aligning with Node APIs where possible, which is why I went with `filename` + an option to disable disk writing. If there's strong objection to this approach and a `filename` + a `filepath` is preferable, I can go with that. Candidly speaking, I think that's a less-obvious API: I don't think it's obvious to consumers that failing to provide a `filepath` would imply no disk writing.

tommoor:
> What's the issue with filename, to stay uniform with Node file system APIs, then an option to disable writing if desired?

Trying to dig into my subconscious to see what's bugging me – I think it's just that inevitably our SDK is going to see heavy use in the browser in the future, and to have to add the `disableWrite` flag in that environment seems odd. Totally agree with your assessment of my previous suggestion though

tommoor:
How about if we detect non-node environment we don't require disableWrite as returning a buffer is the only possible outcome?

bitpshr:
Great idea @tommoor. I added a utility function to detect Node environments and updated all raw save instances to bail if a non-Node environment is detected.

This should be ready for another pass, thanks in advance 🙏

src/types.js
```diff
@@ -75,6 +75,11 @@ export type ListOptions = {
   offset?: number
 };
 
+export type RawOptions = {
+  disableWrite?: boolean,
+  filename?: string
```
evaughn: should `buffer` also be in these options? From https://github.com/goabstract/abstract-sdk/pull/130/files#diff-bfbc29d346bd18b6109a6d4a6e7bfe73R68

Unless I'm misreading the code, then ignore me 😅

src/endpoints/Activities.js
evaughn: > These changes are throughout this PR; we were explicitly specifying a return type when Flow already deduced this correctly for every method we have so far.

Is there an up-sell to removing the explicit return type? Even though Flow correctly infers what the return type is, it might be nice to still include it (for folks who are just jumping into the code for the first time)

src/types.js
bitpshr: A `buffer` is always returned when possible (e.g. raw calls using the API transport), regardless of whether disk writing is enabled. I figured there wasn't much point adding an option to disable that - consumers could just ignore the return value if it's unneeded.

src/endpoints/Activities.js
bitpshr: Normally I agree, but this particular case was really painful to begin with. We're using a generic, and we're having to duplicate this return type in two (sometimes three or four) places in the same method. By removing those that are unnecessary, it really simplifies (and in one place, corrected) the return types.

I'm totally fine with reverting this if there are strong opinions; the main part of this PR is the raw file export functionality, this was just tangential cleanup.

PR #131, opened 2019-05-06
bugfix: add missing LayerDataAsset type

bitpshr:
This pull request brings the `LayerData` type up to parity with core's `LayerData` type. I noticed these discrepancies while working on embeds: I attempted to use a `LayerIcon` component that uses internal functions typed to core's `LayerData` with `LayerData` returned from the SDK.

### DIFF
--- a/docs/abstract-api.md
+++ b/docs/abstract-api.md
@@ -531,6 +531,8 @@ abstract.commits.info({
 |---------------|-------------------|-------------------------------------------------|
 | `childIds`    | `string[]`        | Array of UUID's for the layers children, if any |
 | `id`          | `string`          | UUID of the chid layer                          |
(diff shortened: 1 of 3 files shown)


docs/abstract-api.md
```diff
@@ -531,8 +531,6 @@ abstract.commits.info({
 |---------------|-------------------|-------------------------------------------------|
 | `childIds`    | `string[]`        | Array of UUID's for the layers children, if any |
 | `id`          | `string`          | UUID of the chid layer                          |
-| `libraryId`   | `string`          | UUID of the library file this layer is from     |
-| `libraryName` | `string`          | The name of the library file this layer is from |
```
tommoor: I think the type in `shared` is wrong in this respect, these properties _do_ exist in the data:

image

cc @amccloud

PR #132, opened 2019-05-09
feat(Shares): upgrade to api v13

zpnk:
Shares api v13 automatically resolves the placeholder `{sha: "latest"}` to the layer's _last changed at sha_ `{sha: "abc123"}`.

This is considered a breaking change for any clients depending on the placeholder (this was true for certain internal use cases). 

We'd like to land this in pre-release to coordinate with the internal deprecation of the placeholder.

### DIFF
--- a/src/endpoints/Shares.js
+++ b/src/endpoints/Shares.js
@@ -8,6 +8,10 @@ import type {
 } from "../types";
 import Endpoint from "./Endpoint";
 
+const headers = {
+  "Abstract-Api-Version": "13"
+};
+
 export default class Activities extends Endpoint {
   create<T: Share>(
     descriptor: OrganizationDescriptor,
@@ -21,7 +25,8 @@ export default class Activities extends Endpoint {
             ...descriptor,
             ...shareInput,
             commitSha: (shareInput: any).sha
-          }
+          },
+          headers
         });
       }
     });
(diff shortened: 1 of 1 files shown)


src/endpoints/Shares.js
```diff
@@ -8,6 +8,10 @@ import type {
 } from "../types";
 import Endpoint from "./Endpoint";
 
+const headers = {
+  "Abstract-Api-Version": "13"
```
bitpshr: Should only `Share` requests use the latest API, or should we just bump all API requests to the latest version [here](https://github.com/goabstract/abstract-sdk/blob/master/src/endpoints/Endpoint.js#L223)? Since we’re only planning to move this work out of pre-release once the underlying API is battle-tested, does selective endpoint versioning matter in this case?

zpnk: good question - @amccloud and I discussed bumping the global version and came to the conclusion that it would introduce breaking changes on a number of other endpoints. See the [version list](https://github.com/goabstract/projects/blob/master/stable/api/config/initializers/versions.rb#L15-L19) for some of the changes that have landed since `v8`. Most version additions are due to changes in the response shapes.

We do need to have a discussion around keeping the sdk's version in sync, however for now I think it's fair to upgrade endpoint by endpoint.

Issue #133, opened 2019-05-21
abstract.commits.info() returns latest commit even with sha passed


danielthall:
Regardless of what SHA is passed to this request, the latest commit is always returned:

```
const c = await abstract.commits.info({
    projectId: 'a5161040-9a83-11e8-b10e-77b04fc54df0',
    branchId: 'master',
    sha: 'bf596b716b1993c175a652577c1c956873b478c1'
});
```

responds with:

```
{ sha: 'd47c72c1c4aa42fce84e7caae048f5cbd0b8836c',
  userName: 'Louie Solomon',
  userEmail: '4668cfe9-ccaf-420e-af59-017b2315d394@users.abstractapp.com',
  userId: '4668cfe9-ccaf-420e-af59-017b2315d394',
  title: 'Merged "left nav fix"',
  description: '',
  time: '2019-05-21T06:02:42.000Z',
  parents:
   [ 'bf596b716b1993c175a652577c1c956873b478c1',
     '5cf45717bf029d1b76b841b09d88b8644728a106' ],
  type: 'MERGE',
  fileIds: [],
  sourceBranchId: 'c2d16b97-ad5f-4595-9bbd-7dd8ed63fc4c',
  sourceBranchName: 'left nav fix',
  destinationBranchId: 'master',
  destinationBranchName: 'Master',
  projectId: 'a5161040-9a83-11e8-b10e-77b04fc54df0' }
```

tommoor:
@danielthall thanks for filing, we're looking into it as a matter of priority 👍

bitpshr:
Hi @danielthall, thanks again for taking the time to report this bug. This should be fixed as of `abstract-sdk@v1.5.1`. Please don't hesitate to let us know if you have any additional issues or questions while using the SDK.

danielthall:
Thanks for the quick turn around! Works great!

PR #135, opened 2019-05-21
bugfix: find single commit by sha

bitpshr:
This pull request updates the `commits.info` endpoint to search for and return individual commits by `sha` instead of arbitrarily using the first commit that's returned from `commits.list`.

Fixes https://github.com/goabstract/abstract-sdk/issues/133

### DIFF
--- a/src/endpoints/Commits.js
+++ b/src/endpoints/Commits.js
@@ -13,8 +13,12 @@ export default class Commits extends Endpoint {
   info(descriptor: CommitDescriptor | FileDescriptor | LayerDescriptor) {
     return this.request<Promise<Commit>>({
       api: async () => {
-        const commits = await this.list(descriptor);
-        const commit = commits.find(commit => commit.sha === descriptor.sha);
+        const infoDescriptor = {
+          ...descriptor,
+          startSHA: descriptor.sha
+        };
+
+        const [commit] = await this.list(infoDescriptor, { limit: 1 });
 
         if (!commit) {
           throw new NotFoundError(`sha=${descriptor.sha}`);
--- a/src/endpoints/Commits.js
+++ b/src/endpoints/Commits.js
@@ -13,11 +13,13 @@ export default class Commits extends Endpoint {
   info(descriptor: CommitDescriptor | FileDescriptor | LayerDescriptor) {
     return this.request<Promise<Commit>>({
       api: async () => {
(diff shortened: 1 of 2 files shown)


src/endpoints/Commits.js
```diff
@@ -13,11 +13,13 @@ export default class Commits extends Endpoint {
...
-        if (!commits[0]) {
-          throw new NotFoundError(`commitId=${descriptor.sha}`);
+        const commits = await this.list(descriptor);
+        const commit = commits.find(commit => commit.sha === descriptor.sha);
+
+        if (!commit) {
+          throw new NotFoundError(`sha=${descriptor.sha}`);
         }
-        return commits[0];
+        return commit;
```
bitpshr: This method uses `commits.list` internally because the underlying API doesn't yet support individual commit queries. Since a consumer can query for a commit on a branch / file / layer at any point in time, we need to find all commits for that entity, then iterate over each commit to find one with the `sha` that we're looking for.

For reference, this is how the code [used to work](https://github.com/goabstract/abstract-sdk/blob/6b0133cbdd7576d44c12e4a1dfda8532f6dbf949/src/AbstractAPI/index.js#L296-L305) before the major `v1.0.0` module refactor. We can get away with removing the `descriptor.sha` check now since all supported descriptors have a `sha`.

bitpshr:
Per suggestion from @zpnk, this fix was updated to instead leverage the [`startSha` URL parameter](https://github.com/goabstract/projects/pull/2263) that's now exposed by the API when getting commits for a given entity. This allows the SDK to  always limit result sets to one item and to avoid any iteration or manual searching for the correct `sha`.

src/endpoints/Commits.js
```diff
@@ -13,11 +13,15 @@ export default class Commits extends Endpoint {
   info(descriptor: CommitDescriptor | FileDescriptor | LayerDescriptor) {
     return this.request<Promise<Commit>>({
       api: async () => {
-        const commits = await this.list(descriptor, { limit: 1 });
-        if (!commits[0]) {
-          throw new NotFoundError(`commitId=${descriptor.sha}`);
+        const [commit] = await this.list(descriptor, {
+          limit: 1,
+          startSha: descriptor.sha
```
zpnk: 👍

PR #137, opened 2019-05-31
feature: add pages endpoint to CLI transport

### LINKED ISSUE #134: Pages endpoints in CLI
Currently in the docs we have list/info as being accessible via the CLI, but they're not…
https://sdk.goabstract.com/docs/reference/#pages

Brought to attention here:
https://spectrum.chat/abstract/developers/cli-support-for-pages-info~5c225346-0965-453b-ab26-04f06dc8203a

bitpshr:
This pull request updates the CLI transport by adding support for `Pages#list` and `Pages#info`. It also updates embed documentation and adds this to the sidebar now that this feature is public.

Resolves https://goabstract.atlassian.net/browse/PLATFORM-376
Resolves #134

### DIFF
--- a/docs/embed.md
+++ b/docs/embed.md
@@ -3,9 +3,9 @@ id: embed
 title: Embeds
 ---
 
-Embeds can be used to display interactive, up-to-date previews for layers and collections in any environment that supports HTML. To use an embed, create an `<iframe>` with special URL derived from public share URLs.
+Embeds can be used to display interactive, up-to-date previews for layers and collections in any environment that supports HTML. To use an embed, create an `<iframe>` with a special URL derived from public share URLs.
 
-> Note: Only links for layers that are publicly shared can be embedded.
+> Note: Only links for layers or collections that are publicly shared can be embedded.
 
 ## Creating an embed
 
--- a/docs/embed.md
+++ b/docs/embed.md
@@ -1,11 +1,13 @@
 ---
 id: embed
-title: Embed
+title: Embeds
 ---
 
-> Note: Currently only links for layers that are publicly shared can be embedded
+Embeds can be used to display interactive, up-to-date previews for layers and collections in any environment that supports HTML. To use an embed, create an `<iframe>` with special URL derived from public share URLs.
 
-Embed can be used to display the latest preview for an Abstract layer in your document or application.
+> Note: Only links for layers that are publicly shared can be embedded.
+
+## Creating an embed
 
 Given a public share url:
 
@@ -26,8 +28,4 @@ An embed can be created by using the following URL as the `src` of an `<iframe>`
 ></iframe>
 ```
 
-<iframe src="https://app.goabstract.com/embed/c1182fa5-41d4-4f51-a0e2-5506264b65bd" width="640" height="360" frameborder="0" ></iframe>
-
-<hr />
-
-We will be exploring more themes and options like collections in the near future.
+<iframe src="https://app.goabstract.com/embed/c1182fa5-41d4-4f51-a0e2-5506264b65bd" width="640" height="360" frameborder="0"></iframe>
--- a/src/endpoints/Pages.js
+++ b/src/endpoints/Pages.js
@@ -4,14 +4,28 @@ import { NotFoundError } from "../errors";
 import Endpoint from "./Endpoint";
 
 export default class Pages extends Endpoint {
-  info(descriptor: PageDescriptor) {
+  async info(descriptor: PageDescriptor) {
+    const latestDescriptor = await this.client.descriptors.getLatestDescriptor(
+      descriptor
+    );
+
     return this.request<Promise<Page>>({
       api: async () => {
-        const { pageId, ...fileDescriptor } = descriptor;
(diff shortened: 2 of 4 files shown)


src/endpoints/Pages.js
```diff
@@ -4,14 +4,28 @@ import { NotFoundError } from "../errors";
 import Endpoint from "./Endpoint";
 
 export default class Pages extends Endpoint {
-  info(descriptor: PageDescriptor) {
```
bitpshr: This whole diff is really hard to grep; the main change here was the addition of the `cli` method.

docs/embed.md
```diff
@@ -1,11 +1,13 @@
 ---
 id: embed
-title: Embed
+title: Embeds
 ---
 
-> Note: Currently only links for layers that are publicly shared can be embedded
+Embeds can be used to display interactive, up-to-date previews for layers and collections in any environment that supports HTML. To use an embed, create an `<iframe>` with a special URL derived from public share URLs.
```
zpnk: > To use an embed, create an `<iframe>` with a special URL

Should we say more here? What's the typical flow of getting this url + creating the iframe?

Maybe that info already lives somewhere else, in which case can we link to it?

src/endpoints/Pages.js
```diff
@@ -4,14 +4,28 @@ import { NotFoundError } from "../errors";
...
+    const latestDescriptor = await this.client.descriptors.getLatestDescriptor(
+      descriptor
+    );
+
     return this.request<Promise<Page>>({
       api: async () => {
-        const { pageId, ...fileDescriptor } = descriptor;
+        const { pageId, ...fileDescriptor } = latestDescriptor;
+        const pages = await this.list(fileDescriptor);
+        const page = pages.find(page => page.id === latestDescriptor.pageId);
```
zpnk: total nit: thoughts on using the destructured `pageId` (L14)  here?
```suggestion
        const page = pages.find(page => page.id === pageId);
```
i get that `pageId` would ideally be dropped entirely, but since that's not possible i'd almost rather use it than have an unused var hanging around. surprised eslint doesn't warn in this case.

bitpshr: Good catch.

docs/embed.md
zpnk: lol, ignore this - i just had to scroll down to see the example 😂

src/endpoints/Pages.js
```diff
@@ -4,14 +4,28 @@ import { NotFoundError } from "../errors";
...
+    );
+
     return this.request<Promise<Page>>({
       api: async () => {
-        const { pageId, ...fileDescriptor } = descriptor;
+        const { pageId, ...fileDescriptor } = latestDescriptor;
+        const pages = await this.list(fileDescriptor);
+        const page = pages.find(page => page.id === pageId);
+        if (!page) {
+          throw new NotFoundError(`pageId=${latestDescriptor.pageId}`);
```
zpnk:
```suggestion
          throw new NotFoundError(`pageId=${pageId}`);
```

src/endpoints/Pages.js
```diff
@@ -4,14 +4,28 @@ import { NotFoundError } from "../errors";
...
+        }
+        return page;
+      },
+
+      cli: async () => {
+        const { pageId, ...fileDescriptor } = latestDescriptor;
         const pages = await this.list(fileDescriptor);
-        const page = pages.find(page => page.id === descriptor.pageId);
+        const page = pages.find(page => page.id === latestDescriptor.pageId);
```
zpnk:
```suggestion
        const page = pages.find(page => page.id === pageId);
```

src/endpoints/Pages.js
```diff
@@ -4,14 +4,28 @@ import { NotFoundError } from "../errors";
...
         if (!page) {
-          throw new NotFoundError(`pageId=${descriptor.pageId}`);
+          throw new NotFoundError(`pageId=${latestDescriptor.pageId}`);
```
zpnk:
```suggestion
          throw new NotFoundError(`pageId=${pageId}`);
```

src/endpoints/Pages.js
mattsacks: Could also use `{ id } => id === pageId` 😅

src/endpoints/Pages.js
```diff
@@ -22,15 +36,27 @@ export default class Pages extends Endpoint {
...
+    );
     return this.request<Promise<Page[]>>({
       api: async () => {
         const response = await this.apiRequest(
-          `projects/${descriptor.projectId}/branches/${
-            descriptor.branchId
-          }/files/${descriptor.fileId}/pages`
+          `projects/${latestDescriptor.projectId}/branches/${
+            latestDescriptor.branchId
+          }/files/${latestDescriptor.fileId}/pages`
```
mattsacks: 🧶 – Might make this a bit cleaner to take all of these properties out of `latestDescriptor`:


```js
const { branchId, fileId, projectId } = latestDescriptor;
const response = await this.apiRequest(
  `projects/${projectId}/branches/${branchId}/files/${fileId}/pages`
);
```

PR #142, opened 2019-06-27
test commit message

bitpshr:
This pull request is only meant to test integration with the [Semantic Pull Requests](https://probot.github.io/apps/semantic-pull-requests/) bot.


bitpshr:
Update: Integration working as expected. I really like not enforcing `commitlint` on the per-commit level as part of CI; good call @tommoor.

PR #145, opened 2019-07-24
docs: Document the `updatedAt` field in the File object

huw:
I noticed the File object now comes with a (very useful) `updatedAt` field. I'm raising this to document it in the APIs.


bitpshr:
Good catch. Thanks very much for the contribution, @huw!

PR #146, opened 2019-07-24
docs: Fix a typo in the Files API documentation

huw:
`Retrieve an Sketch file` → `Retrieve a Sketch file`.

Note that this will break any permalinks to the header.


bitpshr:
Another good catch. Thanks for taking the time to fix this.

PR #149, opened 2019-07-30
fix: commits.list input descriptor and docs

rickharris:
I tried to use `commits.list` inside of abstract and couldn't do it the way it's currently typed, and I found some types that need updating. Will leave comments inline.

### DIFF
--- a/docs/abstract-api.md
+++ b/docs/abstract-api.md
@@ -520,9 +520,7 @@ abstract.commits.list({
   projectId: "616daa90-1736-11e8-b8b0-8d1fec7aef78",
   branchId: "master",
   fileId: "51DE7CD1-ECDC-473C-B30E-62AE913743B7",
-  pageId: "7D2D2599-9B3F-49BC-9F86-9D9D532F143A",
   layerId: "CA420E64-08D0-4B96-B0F7-75AA316B6A19",
-  sha: "latest"
 });
 ```
 
--- a/docs/abstract-api.md
+++ b/docs/abstract-api.md
@@ -502,7 +502,7 @@ to identify which version of the object you would like.
 
 ![CLI][cli-icon] ![API][api-icon]
 
-`commits.list(BranchDescriptor | LayerDescriptor): Promise<Commit[]>`
+`commits.list(BranchDescriptor, options): Promise<Commit[]>`
 
 List the commits for a specific branch
 
@@ -513,14 +513,18 @@ abstract.commits.list({
 });
 ```
 
-or, get a list of commits for a layer – this query will only return commits where the referenced layer was changed…
+or, filter the list of branch commits to a single layer by providing a file id and layer id. This will only return commits where the referenced layer was changed…
 
 ```js
-abstract.commits.list({
-  projectId: "616daa90-1736-11e8-b8b0-8d1fec7aef78",
-  branchId: "master",
-  fileId: "51DE7CD1-ECDC-473C-B30E-62AE913743B7",
-  layerId: "CA420E64-08D0-4B96-B0F7-75AA316B6A19",
+abstract.commits.list(
+  {
+    projectId: "616daa90-1736-11e8-b8b0-8d1fec7aef78",
+    branchId: "master",
+  },
+  {
+    fileId: "51DE7CD1-ECDC-473C-B30E-62AE913743B7",
+    layerId: "CA420E64-08D0-4B96-B0F7-75AA316B6A19"
+  }
 });
 ```
 
--- a/src/endpoints/Commits.js
+++ b/src/endpoints/Commits.js
@@ -45,15 +45,16 @@ export default class Commits extends Endpoint {
   ) {
     return this.request<Promise<Commit[]>>({
       api: async () => {
-        const query = querystring.stringify({
+        let query = querystring.stringify({
           fileId: options.fileId,
           layerId: options.layerId,
           startSHA: options.startSHA,
           endSHA: options.endSHA,
           limit: options.limit
         });
+        query = query ? `?${query}` : "";
         const response = await this.apiRequest(
-          `projects/${descriptor.projectId}/branches/${descriptor.branchId}/commits?${query}`
+          `projects/${descriptor.projectId}/branches/${descriptor.branchId}/commits${query}`
         );
         return response.commits;
       },
@@ -65,8 +66,8 @@ export default class Commits extends Endpoint {
           descriptor.branchId,
           ...(options.fileId ? ["--file-id", options.fileId] : []),
           ...(options.layerId ? ["--layer-id", options.layerId] : []),
-          ...(options.startSHA ? ["--file-id", options.startSHA] : []),
-          ...(options.endSHA ? ["--layer-id", options.endSHA] : []),
+          ...(options.startSHA ? ["--start-sha", options.startSHA] : []),
+          ...(options.endSHA ? ["--end-sha", options.endSHA] : []),
           ...(options.limit ? ["--limit", options.limit.toString()] : [])
         ]);
         return response.commits;
--- a/src/endpoints/Commits.js
+++ b/src/endpoints/Commits.js
@@ -1,6 +1,7 @@
 // @flow
 import querystring from "query-string";
 import type {
+  BranchDescriptor,
   Commit,
   CommitDescriptor,
   FileDescriptor,
@@ -33,7 +34,7 @@ export default class Commits extends Endpoint {
   }
 
   list(
-    descriptor: CommitDescriptor | FileDescriptor | LayerDescriptor,
+    descriptor: BranchDescriptor | LayerDescriptor,
     options: { limit?: number } = {}
   ) {
     return this.request<Promise<Commit[]>>({
--- a/src/endpoints/Commits.js
+++ b/src/endpoints/Commits.js
@@ -34,15 +34,23 @@ export default class Commits extends Endpoint {
   }
 
   list(
-    descriptor: BranchDescriptor | LayerDescriptor,
-    options: { limit?: number } = {}
+    descriptor: BranchDescriptor,
+    options: {
+      fileId?: string,
+      layerId?: string,
+      startSHA?: string,
+      endSHA?: string,
+      limit?: number
+    } = {}
   ) {
     return this.request<Promise<Commit[]>>({
       api: async () => {
         const query = querystring.stringify({
-          ...options,
-          fileId: descriptor.fileId && descriptor.fileId,
-          layerId: descriptor.layerId && descriptor.layerId
+          fileId: options.fileId,
+          layerId: options.layerId,
+          startSHA: options.startSHA,
+          endSHA: options.endSHA,
+          limit: options.limit
         });
         const response = await this.apiRequest(
           `projects/${descriptor.projectId}/branches/${descriptor.branchId}/commits?${query}`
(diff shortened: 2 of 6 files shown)


src/endpoints/Commits.js
```diff
@@ -33,15 +34,23 @@ export default class Commits extends Endpoint {
   }
 
   list(
-    descriptor: CommitDescriptor | FileDescriptor | LayerDescriptor,
-    options: { limit?: number } = {}
+    descriptor: BranchDescriptor,
+    options: {
+      fileId?: string,
+      layerId?: string,
```
rickharris: 🍖 There are a couple of changes represented here:

- `descriptor: CommitDescriptor | FileDescriptor | LayerDescriptor,` I believe was a copy/paste error and was intended to be `BranchDescriptor | LayerDescriptor`. Neither api or cli support `CommitDescriptor` or `FileDescriptor` for commit history.
- After I made that change, I realized that `LayerDescriptor` doesn't make much sense either, because it includes `sha`, which is not used in calculating the commit history of a layer. It might make sense if you wanted to see the commit history starting with a particular layer at a particular sha as described by the descriptor, but what this endpoint is actually doing is returning branch history filtered down to a specific `fileId` and `layerId` combination. Moving `fileId` and `layerId` to `options` more accurately reflects the "branch history, filtered" idea, and is more accurate to what's going on behind the scenes and doesn't require a `sha` to be passed in when it is unnecessary and maybe a little confusing when it's ignored.

src/endpoints/Commits.js
```diff
@@ -33,15 +34,23 @@ export default class Commits extends Endpoint {
...
+      startSHA?: string,
+      endSHA?: string,
```
rickharris: These are valid options in both api and cli.

src/endpoints/Commits.js
```diff
@@ -33,15 +34,23 @@ export default class Commits extends Endpoint {
...
       api: async () => {
         const query = querystring.stringify({
-          ...options,
-          fileId: descriptor.fileId && descriptor.fileId,
-          layerId: descriptor.layerId && descriptor.layerId
+          fileId: options.fileId,
+          layerId: options.layerId,
+          startSHA: options.startSHA,
+          endSHA: options.endSHA,
+          limit: options.limit
```
rickharris: We were already filtering by valid, known options in the cli transport so I figured I'd do it here as well.

src/endpoints/Commits.js
```diff
@@ -54,8 +63,10 @@ export default class Commits extends Endpoint {
           "commits",
           descriptor.projectId,
           descriptor.branchId,
-          ...(descriptor.fileId ? ["--file-id", descriptor.fileId] : []),
-          ...(descriptor.layerId ? ["--layer-id", descriptor.layerId] : []),
+          ...(options.fileId ? ["--file-id", options.fileId] : []),
+          ...(options.layerId ? ["--layer-id", options.layerId] : []),
+          ...(options.startSHA ? ["--file-id", options.startSHA] : []),
+          ...(options.endSHA ? ["--layer-id", options.endSHA] : []),
```
rickharris: add support for `startSHA` and `endSHA` to cli transport. These were already kind of supported in api because of the `...options` spread, so I wanted to make it available here also.

rickharris:
CI is failing because of a decrease in coverage - is there a way to see the coverage areas to find out what in this PR is uncovered currently?

src/endpoints/Commits.js
tommoor: > Moving fileId and layerId to options more accurately reflects the "branch history, filtered" idea, and is more accurate to what's going on behind the scenes

💯

src/endpoints/Commits.js
```diff
@@ -54,8 +63,10 @@ export default class Commits extends Endpoint {
           "commits",
           descriptor.projectId,
           descriptor.branchId,
-          ...(descriptor.fileId ? ["--file-id", descriptor.fileId] : []),
-          ...(descriptor.layerId ? ["--layer-id", descriptor.layerId] : []),
+          ...(options.fileId ? ["--file-id", options.fileId] : []),
+          ...(options.layerId ? ["--layer-id", options.layerId] : []),
+          ...(options.startSHA ? ["--file-id", options.startSHA] : []),
```
tommoor:
```suggestion
          ...(options.startSHA ? ["--start-sha", options.startSHA] : []),
```

src/endpoints/Commits.js
tommoor:
```suggestion
          ...(options.endSHA ? ["--end-sha", options.endSHA] : []),
```

bitpshr:
@rickharris: You should have a `coverage` directory that's output during testing; you can serve this locally to view a static `lcov` report.

docs/abstract-api.md
```diff
@@ -502,7 +502,7 @@ to identify which version of the object you would like.
 
 ![CLI][cli-icon] ![API][api-icon]
 
-`commits.list(BranchDescriptor | LayerDescriptor): Promise<Commit[]>`
```
rickharris: Any suggestion for documenting `options` in a nice way formatting-wise?

src/endpoints/Files.js
```diff
@@ -16,8 +16,8 @@ export default class Files extends Endpoint {
     );
     return this.request<Promise<File>>({
       api: async () => {
-        const { fileId, ...branchDescriptor } = latestDescriptor;
-        const files = await this.list(branchDescriptor);
+        const { fileId, ...commitDescriptor } = latestDescriptor;
+        const files = await this.list(commitDescriptor);
```
rickharris: Random change found while hunting down code coverage - this `...` here includes a `sha` and `this.list` takes `CommitDescriptor` so I renamed this variable to match its type.

docs/abstract-api.md
bitpshr: Not really; our API documentation is especially manual since there's still no viable documentation solution that can infer types properly from Flow. Defining an `options` type separately here for clarity or just documenting it inline should both work for now.

amccloud: I'm sorta confused with this change and sorry if I missed the explanation. Are we not able to list commits for a layer anymore? The idea with descriptors was to make a clear distinction between identifiers used to reference objects and options used to modify behavior

### List commits for branch
```js
abstract.commits.list(branchDescriptor);
```

### List commits for file
```js
abstract.commits.list(fileDescriptor); // branchDescriptor + fileId
```


### List commits for layer
```js
abstract.commits.list(laterDescriptor); // fileDescriptor + layerId
```

rickharris: Explanation about descriptors here https://github.com/goabstract/abstract-sdk/pull/149#discussion_r308545630

amccloud: @rickharris so I think another fix is to make `sha` on `LayerDescrpitor` optional or add a version of `LatestLayerDescriptor` with sha so we have one with and without sha. Technically `sha: "latest"` is the same as omitting the sha (optional sha was originally how it worked) cc: @joesteele 

When you do pass a sha to a `LayerDescriptor` I'd expect to get the history starting from the sha.

This change sorta breaks the model of the entire SDK

```js
client.type.action(object, options)
```

object can be anything and we refer to it with a descriptor. In the context of data in git data the object can have multiple versions that you can access by selecting the version you want with a sha. We always require the  sha right now though, but you can get the latest version with `sha: "latest"`, which as I mentioned earlier would be like omitting the sha and just taking the latest object

rickharris: Like I pointed out in the other comment thread, thinking of this as "history of a layer" is not exactly accurate, it's more like "branch history, filtered to a specific layer id". I don't think it really breaks the model of the sdk, you're requesting commit history for a branch, which you pass as the descriptor, filtered to a specific layer id, which you pass in the `options`.

rickharris: Said another way, let's say that we support passing `LayerDescriptor` into `commits.list`, the commit list would still be limited to a single branch, which might not be obvious if you were expecting a full layer history for the given descriptor. The list of commits would just unceremoniously stop at the "branched from" commit.

bitpshr: I do think that where possible, we should abstract how the API or CLI works under the hood and holistically use descriptors to represent singular query sets, like a commits list in this case. Descriptors representing query sets is one of the reasons we have so many specialized descriptors already instead of combining them and making certain properties optional.

I don't have strong feelings either way, but I think that adding a new descriptor to fit this use-case is more in-line with our API than passing in secondary options like @amccloud suggested.

amccloud: > I don't think it really breaks the model of the sdk

The idea is you can do this

```js
const layerDescriptor = layerDescriptorFromLocation(location);
const preview = await abstrat.previews.url(layerDescriptor);
const project = await abstrat.projects.info(layerDescriptor);
const commits = await abstrat.commits.list(layerDescriptor);
```

with this change, it'll be:

```js
const layerDescriptor = layerDescriptorFromLocation(location);
const preview = await abstrat.previews.url(layerDescriptor);
const project = await abstrat.projects.info(layerDescriptor);
const commits = await abstrat.commits.list(branchDesriptor(layerDescriptor), {
  layerId: layerDescriptor.layerId
});

// branchDesriptor helper (expected) is used to coerce layerDescriptor into exact branch descriptor
```

> Like I pointed out in the other comment thread, thinking of this as "history of a layer" is not exactly accurate, it's more like "branch history, filtered to a specific layer id". I don't think it really breaks the model of the sdk, you're requesting commit history for a branch, which you pass as the descriptor, filtered to a specific layer id, which you pass in the options.

I noted that, but that feels more like an implementation detail. Are the commits returned related to the layer? Another lens to maybe view it from is that it's just an in-memory join. If you compare to something like a rails relation: `organization.users`, that's technically all users filtered down by organizationId or from another lens users that belong to the org

> which might not be obvious if you were expecting a full layer history for the given descriptor. The list of commits would just unceremoniously stop at the "branched from" commit.

I don't think we want to stop at "branched from" for much longer. This has been also confusing in [other context](https://goabstract.atlassian.net/browse/CORE-472?filter=-2), not just the SDK. Controlling this behavior seems like an option - just like limit or any of the other options we use to decide where to stop

tommoor: @amccloud I totally see where you're coming from, the example above is a much nicer API in the first version. I think where it falls apart is in the case of manually constructing the descriptor (like Rick is doing, and we are in the desktop app from route params) – and there is literally no `sha` available. As an externally exposed API what would you suggest to end users, they add a "junk" `sha` parameter in to make it layerDescriptor compatible?

amccloud: @tommoor yup, the absence of sha would be `sha: "latest"`. At first, layerDescriptor was typed as:

```js
{..., layerId: string, sha?: string}
```

but then we went with

```js
{..., layerId: string, sha: string | "latest"}
```

I don't think I made a good case for omitting sha previously to mean latest but this was why. It just worked as you expected in more scenarios

@joesteele ran into a similar problem and had to introduce new sha-less types to address it. My recommendation would be to update the sdk to make `sha` optional, and maybe add `LayerVersionDescriptor` or something along those lines if we find ourselves in a scenario where sha is 100% required. In that case it would be likely the code revolves around specific versions of the layer and not the layer

bitpshr:
Thanks again for this @rickharris. I took the core fix here and reworked it with the suggestion from @amccloud as part of https://github.com/goabstract/abstract-sdk/pull/171.

PR #151, opened 2019-07-30
fix: disable caching of non-static entities

bitpshr:
This pull request removes caching from non-static endpoints. This was a total oversight and could result in stale, cached entities being returned instead of the most up-to-date version.

After this change, the entity types below are the only ones that we'll cache (since they are all either static or requested using a static `sha`):

```
Activity
Asset
Changeset
Commit
Data
File
Layer
Membership
Notification
Preview
Share
```

References https://goabstract.atlassian.net/browse/PLATFORM-490

### DIFF
--- a/src/endpoints/Branches.js
+++ b/src/endpoints/Branches.js
@@ -19,10 +19,6 @@ export default class Branches extends Endpoint {
           descriptor.projectId,
           descriptor.branchId
         ]);
-      },
-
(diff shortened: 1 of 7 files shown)


bitpshr:
@amccloud agreed; we could couple that with the ability to explicitly disable caching during a request by adding an option that all cached endpoints support for max-portability.

PR #157, opened 2019-08-19
Typescript 3.5.3 type errors fixed

laitine:
Typescript compilation fails due to 3 errors in the type file abstract-sdk.d.ts. Tested on Typescript versions 3.5.2 and 3.5.3.

### DIFF
--- a/abstract-sdk.d.ts
+++ b/abstract-sdk.d.ts
@@ -198,7 +198,7 @@ interface Commits extends Endpoint {
 
   list(
     descriptor: CommitDescriptor | FileDescriptor | LayerDescriptor,
-    options: { limit?: number }
+    options?: { limit?: number }
   ): Promise<Commit[]>;
 }
 
--- a/abstract-sdk.d.ts
+++ b/abstract-sdk.d.ts
@@ -104,7 +104,7 @@ interface Endpoint {
   transportMode: string;
   webUrl: string | Promise<string>;
 
-  constructor(client: Client, options: CommandOptions): any;
+  constructor(client: Client, options: CommandOptions): Endpoint;
   accessToken: () => Promise<AccessToken>;
   cliRequest(args: string[]): Promise<any>;
   request<T>(request: EndpointRequest<T>): T;
@@ -1332,9 +1332,9 @@ type LayerTextStyle = {
 };
 
 type LayerOverrideData = {
-  symbolId?: any,
+  symbolId?: string,
   properties?: any,
-  [layerId: string]: LayerOverrideData
+  [layerId: string]: string | undefined
 };
 
 type LayerDataAsset = {
--- a/abstract-sdk.d.ts
+++ b/abstract-sdk.d.ts
@@ -104,7 +104,7 @@ interface Endpoint {
   transportMode: string;
   webUrl: string | Promise<string>;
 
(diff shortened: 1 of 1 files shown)


bitpshr:
Thanks for submitting this @laitine! Could you please share the error output that caused you to notice these errors? I just want to fully understand the issues that these changes are meant to fix.

Thanks in advance.

laitine:
$ tsc -p 'scripts'
node_modules/abstract-sdk/abstract-sdk.d.ts:107:3 - error TS7010: 'constructor', which lacks return-type annotation, implicitly has an 'any' return type.

107   constructor(client: Client, options: CommandOptions);
      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

node_modules/abstract-sdk/abstract-sdk.d.ts:201:5 - error TS2371: A parameter initializer is only allowed in a function or constructor implementation.

201     options: { limit?: number } = {}
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

node_modules/abstract-sdk/abstract-sdk.d.ts:1335:3 - error TS2411: Property 'symbolId' of type 'string | undefined' is not assignable to string index type 'LayerOverrideData'.

1335   symbolId?: string,
       ~~~~~~~~


Found 3 errors.

abstract-sdk.d.ts
```diff
@@ -104,7 +104,7 @@ interface Endpoint {
   transportMode: string;
   webUrl: string | Promise<string>;
 
-  constructor(client: Client, options: CommandOptions);
+  constructor(client: Client, options: CommandOptions): any;
```
bitpshr:
```suggestion
  constructor(client: Client, options: CommandOptions): Endpoint;
```

abstract-sdk.d.ts
```diff
@@ -198,7 +198,7 @@ interface Commits extends Endpoint {
 
   list(
     descriptor: CommitDescriptor | FileDescriptor | LayerDescriptor,
-    options: { limit?: number } = {}
+    options: { limit?: number }
```
bitpshr:
```suggestion
    options?: { limit?: number }
```

abstract-sdk.d.ts
```diff
@@ -1332,7 +1332,7 @@ type LayerTextStyle = {
 };
 
 type LayerOverrideData = {
-  symbolId?: string,
+  symbolId?: any,
```
bitpshr: This should stay typed as a `string`. If you're getting an error about `undefined | string` not being assignable to a `string`, first check if the value is actually a string and the error should go away.

laitine: The issue is caused by Typescript forcing proper index type as string for objects. Gives the following error "TS2411: Property 'symbolId' of type 'string | undefined' is not assignable to string index type 'string'." It compiles with the following:

```javascript
type LayerOverrideData = {
  symbolId?: string,
  properties?: any,
  [layerId: string]: string | undefined
};
```

bitpshr:
Thanks again for this work and for identifying these issues @laitine. I incorporated most of your changes in 501ee159c9364abc498db84d59df6c8230b97e4a, which will be included in our next release planned for later this week.

PR #166, opened 2019-10-25
chore: upgrade flow to 110

rickharris:
Bumps flow-bin to v0.110.1 and fixes one resulting flow error, which seems like it won't cause any issues in projects that depend on abstract-sdk.

### DIFF
--- a/src/endpoints/Endpoint.js
+++ b/src/endpoints/Endpoint.js
@@ -198,7 +198,7 @@ export default class Endpoint {
     return response;
   }
 
-  async _getAPIHeaders(headers?: {}) {
+  async _getAPIHeaders(customHeaders?: { [key: string]: string }) {
     let tokenHeader = {};
     const token = await this.accessToken();
 
@@ -216,18 +216,23 @@ export default class Endpoint {
           : { "Abstract-Share-Id": inferShareId(token) };
     }
 
-    const tokens: { [string]: ?string } = {
+    const baseHeaders = {
       Accept: "application/json",
       "Content-Type": "application/json",
       "User-Agent": `Abstract SDK ${minorVersion}`,
       "X-Amzn-Trace-Id": `Root=1-${new Date().getTime()}-${uuid()}`,
-      "Abstract-Api-Version": "8",
-      ...tokenHeader,
-      ...headers
+      "Abstract-Api-Version": "8"
     };
-    Object.keys(tokens).forEach(key => {
-      tokens[key] === undefined && delete tokens[key];
+
+    const headers = {
+      ...baseHeaders,
+      ...tokenHeader,
+      ...customHeaders
+    };
+
+    Object.keys(headers).forEach(key => {
+      headers[key] === undefined && delete headers[key];
     });
-    return tokens;
+    return headers;
   }
 }
--- a/src/endpoints/Endpoint.js
+++ b/src/endpoints/Endpoint.js
@@ -216,7 +216,7 @@ export default class Endpoint {
           : { "Abstract-Share-Id": inferShareId(token) };
     }
 
-    const tokens = {
+    const tokens: { [string]: ?string } = {
       Accept: "application/json",
       "Content-Type": "application/json",
       "User-Agent": `Abstract SDK ${minorVersion}`,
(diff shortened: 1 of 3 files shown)


src/endpoints/Endpoint.js
```diff
@@ -216,7 +216,7 @@ export default class Endpoint {
           : { "Abstract-Share-Id": inferShareId(token) };
     }
 
-    const tokens = {
+    const tokens: { [string]: ?string } = {
```
rickharris: Fixed this error:

```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/endpoints/Endpoint.js:229:43

Cannot delete tokens[key] because:
 • undefined [1] is incompatible with string [2].
 • undefined [1] is incompatible with string [3].
 • undefined [1] is incompatible with string [4].
 • undefined [1] is incompatible with string [5].
 • undefined [1] is incompatible with string [6].

 [2] 220│       Accept: "application/json",
 [3] 221│       "Content-Type": "application/json",
 [4] 222│       "User-Agent": `Abstract SDK ${minorVersion}`,
 [5] 223│       "X-Amzn-Trace-Id": `Root=1-${new Date().getTime()}-${uuid()}`,
 [6] 224│       "Abstract-Api-Version": "8",
     225│       ...tokenHeader,
     226│       ...headers
     227│     };
     228│     Object.keys(tokens).forEach(key => {
 [1] 229│       tokens[key] === undefined && delete tokens[key];
     230│     });
     231│     return tokens;
     232│   }
```

amccloud: At a quick glance the error seems valid. Why will `key === undefined`? It also looks like `tokens` really should be called `headers`

I think this annotation can be moved up to `headers` argument (maybe as  `customHeaders: { [key: string]: string | void }`) 

Using `void` over `?` to disallow `null` headers

rickharris: @amccloud To be honest I don't understand this function at all so I didn't want to rename things I didn't understand 😂. And I agree that the error is at least semi-valid because this is what flow thinks `tokens` is if you don't give it a type:

```
{|
  "Abstract-Api-Version": string, 
  "Abstract-Share-Id": (void | string), 
  Accept: string, 
  Authorization: (void | string), 
  "Content-Type": string, 
  "User-Agent": string, 
  "X-Amzn-Trace-Id": string
|}
```

So if you `delete tokens["User-Agent"]` it doesn't like that, since `User-Agent` is hard-coded as `string` when defining `tokens`. That's why I gave `tokens` in particular the type, and why I don't think it would have the same affect if we typed the `headers` argument instead.

amccloud: ✅ my suggestion does not work as expected https://flow.org/try/#0C4TwDgpgBAwgrgZ2AewLYAkIEMAmEBOCUAvFAN5QDaA1hCAFxRL4CWAdgOYC6jz7HUAD5QAbshY4oAXwDcAKFCQoAFWS02RUmUFyoUAEQBBAEbMsAY2ABaQ2BZWAagQQtkbfb2CtOAGii6DEzNLKwBlAAssfAgrAEkcDygACjEJISYvfgBKPwDDc3MIMGBPbw5cvUM4YHDkVgAvLGBXNkYU8UlhPk4c-z19GDdgCDZrZXAIRO7yvoMAVQQCGw4R4CnM31n9AA0bVHq2K2V8Cxj49bKfAIA6W-gkNExcZzlBWTk5PHMAGyjoESiUBQ6gQjFUIPknwg3wgwyBahGCEoAHIFktDCtRsiuPIgA

Pretty sure on the variable name suggestion though! These are the base request headers for the sdk and this spread allows overriding or customizing the headers. If you send a custom header of `undefined` we unset was conflicting with the inferred exact type you shared above

Moving the base headers into their own object might be another workaround:

```js
const tokens = {...baseHeaders, ...tokenHeaders, ...overrideHeaders}; // avoid the exactness and everything should allow optional without an explicit annotation
````

rickharris: @amccloud I some changes in https://github.com/goabstract/abstract-sdk/commit/48fe5710a7074f4bbe33e2d656665bec277855e2, lmk what you think

PR #171, opened 2019-11-22
fix: update commits.list method signature

bitpshr:
This pull request fixes the `commits.list` method by updating its signature accordingly. To help accomplish this, the existing `LayerDescriptor` was updated to be `sha`-less, and a new `LayerVersionDescriptor` was added for cases where a `sha` is actually required, which is still most cases.

### DIFF
--- a/abstract-sdk.d.ts
+++ b/abstract-sdk.d.ts
@@ -201,7 +201,7 @@ interface Comments extends Endpoint {
       BranchDescriptor
       | CommitDescriptor
       | PageDescriptor
-      | LayerDescriptor & { pageId: string },
+      | LayerVersionDescriptor & { pageId: string },
     comment: NewComment
   ): Promise<Comment>;
 
@@ -209,7 +209,7 @@ interface Comments extends Endpoint {
     descriptor:
       BranchDescriptor
       | CommitDescriptor
-      | LayerDescriptor
+      | LayerVersionDescriptor
       | PageDescriptor,
     options?: ListOptions
   ): CursorPromise<Comment[]>;
@@ -217,17 +217,21 @@ interface Comments extends Endpoint {
 
 interface Commits extends Endpoint {
   info(
-    descriptor: CommitDescriptor | FileDescriptor | LayerDescriptor
+    descriptor: CommitDescriptor | FileDescriptor | LayerVersionDescriptor
   ): Promise<Commit>;
 
   list(
-    descriptor: CommitDescriptor | FileDescriptor | LayerDescriptor,
-    options?: { limit?: number }
+    descriptor: BranchDescriptor | FileDescriptor | LayerDescriptor,
+    options?: {
+      limit?: number,
+      startSHA?: string,
+      endSHA?: string
+    }
   ): Promise<Commit[]>;
 }
 
 interface Data extends Endpoint {
-  info(descriptor: LayerDescriptor): Promise<LayerDataset>;
+  info(descriptor: LayerVersionDescriptor): Promise<LayerDataset>;
 }
 
 interface Descriptors extends Endpoint {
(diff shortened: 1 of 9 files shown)


src/types.js
```diff
@@ -59,6 +59,14 @@ export type PageDescriptor = {|
...
 export type LayerDescriptor = {|
+  sha?: "latest" | string,
+  projectId: string,
+  branchId: string | "master",
+  fileId: string,
+  layerId: string
+|};
+
+export type LayerVersionDescriptor = {|
   ...$Exact<ObjectDescriptor>,
```
amccloud: nit: Could technically spread LayerDescriptor and make sha non-optional

abstract-sdk.d.ts
```diff
@@ -201,33 +201,37 @@ interface Comments extends Endpoint {
...
 
   list(
-    descriptor: CommitDescriptor | FileDescriptor | LayerDescriptor,
-    options?: { limit?: number }
+    descriptor: BranchDescriptor | FileDescriptor | LayerDescriptor,
+    options?: {
+      limit?: number,
+      startSHA?: string,
+      endSHA?: string
+    }
```
bitpshr: This is the main change in this PR that fixes the method signature. All other changes are churn from `LayerDescriptor` and `LayerVersionDescriptor` changes.

tommoor: Context: I ran into this attempting to use the SDK in the desktop app:

image

abstract-sdk.d.ts
```diff
@@ -788,7 +800,7 @@ type PageShareInput = BaseShareInput & PageDescriptor & {
   kind: "page"
 };
 
-type LayerShareInput = BaseShareInput & LayerDescriptor & {
+type LayerShareInput = BaseShareInput & LayerVersionDescriptor & {
```
amccloud: Any way to get this as a concrete object in typescript? These intersections don't seem ideal

tommoor:
@bitpshr does this supercede https://github.com/goabstract/abstract-sdk/pull/149 ?

rickharris:
Pretty sure it does. Wow I really lost track of that PR somehow, I'm not even sure I had seen all the discussion on it.

bitpshr:
Yup, this work does supersede #149 (just closed it for clarity.)

src/endpoints/Commits.js
```diff
@@ -33,8 +35,12 @@ export default class Commits extends Endpoint {
   }
 
   list(
-    descriptor: CommitDescriptor | FileDescriptor | LayerDescriptor,
-    options: { limit?: number } = {}
+    descriptor: BranchDescriptor | FileDescriptor | LayerDescriptor,
```
rickharris: From https://github.com/goabstract/abstract-sdk/pull/149/files#r308545630, I think we can remove `FileDescriptor` here.

src/types.js
```diff
@@ -59,6 +59,14 @@ export type PageDescriptor = {|
 |};
 
 export type LayerDescriptor = {|
+  sha?: "latest" | string,
```
rickharris: Why do we still want this here as optional? Would it ever get used if someone passed it?

abstract-sdk.d.ts
bitpshr: Sorry, missed this comment last week. Why are intersection types not ideal in this case?

amccloud: Likely not a huge deal in this case. I've avoided intersections because they are often used to create types on the fly and display:

`BaseShareInput & LayerVersionDescriptor & {...}` 

in tools vs just:

`LayerShareInput`

or something more concrete

bitpshr: Ah, interesting. I never noticed the autocomplete issue with intersection types. I will rework those in a followup PR for sure.

bitpshr:
Since all feedback so far had been addressed, I went ahead and landed this for the sake of getting a patched `commits.list` method signature out the door. If there's any additional feedback, I will definitely take care of it in a followup PR.

PR #172, opened 2019-12-06
feat: multi transport support

bitpshr:
This pull request adds multi-transport support to the SDK.

## Breaking changes

- `transportMode` is now an array of strings, e.g. `["cli", "api"]`.
- The specified transports will be tried sequentially until one succeeds.
- If no transport succeeds, a `MultiError` will be thrown with references to all transport errors.
- Every SDK endpoint now accepts an options argument that can be used to specify a one-off `transportMode` that overrides the globally-configured `transportMode`.

## Internal imrpovements

Several internal improvements were made along the way, including:

- `Endpoint` was rewritten and is now fully tested.
- Module aliases were added.
- `CursorPromise` was simplified.
- Unused internal caching was removed.
- Folder structure was consolidated.
- Eslint and Jest errors were fixed.


tests/Client.test.js
```diff
@@ -0,0 +1,178 @@
...
+  MultiError,
+  NotFoundError
+} from "@core/errors";
+import { // eslint-disable-line
+  mockAPI,
+  mockCLI,
+  API_CLIENT,
+  CLIENT_CONFIG
+} from "@core/util/testing";
+import Client from "@core/Client"; // eslint-disable-line
```
bitpshr: Eslint is disabled for these lines to prevent import reordering.

package.json
```diff
@@ -84,11 +86,22 @@
     "extends": "@elasticprojects/eslint-config-abstract",
     "env": {
       "jest": true
+    },
+    "settings": {
+      "react": {
+        "version": "^16.0.0"
```
bitpshr: This is necessary to silence warnings from `@elasticprojects/eslint-config-abstract`.

PR #173, opened 2019-12-12
feat: Remove need to pass branchId to load a commit

tommoor:
BREAKING CHANGES

Depends on deployment of new API endpoint: https://github.com/goabstract/projects/pull/3735

### DIFF
--- a/docs/abstract-api.md
+++ b/docs/abstract-api.md
@@ -1684,6 +1684,15 @@ Reference for the parameters required to load resources with the Abstract SDK.
 }
 ```
 
+### CommitDescriptor
+
+```js
+{
+  projectId: string,
+  branchId: string | "master"
+}
+```
+
 ### FileDescriptor
 
 ```js
--- a/docs/abstract-api.md
+++ b/docs/abstract-api.md
@@ -121,7 +121,7 @@ abstract.assets.file({
 
 ### List assets for a commit
 
-`assets.commit(CommitDescriptor, RequestOptions): Promise<Asset[]>`
+`assets.commit(BranchCommitDescriptor, RequestOptions): Promise<Asset[]>`
 
 List all assets generated for a commit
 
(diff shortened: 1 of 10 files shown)


src/endpoints/Commits.js
```diff
@@ -19,9 +19,15 @@ export default class Commits extends Endpoint {
     return this.configureRequest<Promise<Commit>>(
       {
         api: () => {
-          return this.apiRequest(
-            `projects/${descriptor.projectId}/branches/${descriptor.branchId}/commits/${descriptor.sha}`
-          );
+          // loading commits with a share token requires a branchId so this
+          // route is maintained for that circumstance
+          return descriptor.branchId
```
tommoor: 🍖

docs/abstract-api.md
```diff
@@ -1675,7 +1674,7 @@ Reference for the parameters required to load resources with the Abstract SDK.
 }
 ```
 
-### CommitDescriptor
+### BranchCommitDescriptor
```
bitpshr: We should also add a new `CommitDescriptor` reference here that shows a descriptor without a `branchId`.

src/endpoints/Commits.js
bitpshr: I like that we're maintaining the old route here, but I think that we need to update this method signature to also support `BranchCommitDescriptor` in addition to `CommitDescriptor` (see my comment below re: removing `branchId` altogether from `CommitDescriptor` rather than making optional.)

Unrelated to the work in this PR, I wonder why we support `FileDescriptor` and `LayerVersionDescriptor` despite not using anything other than `projectId`, `branchId`, and `sha`. Maybe it's just for convenience so new, sub-descriptors don't need to be created before hitting this endpoint, but that feels weird...

src/types.js
```diff
@@ -37,12 +37,18 @@ export type ObjectDescriptor = {
...
+export type BranchCommitDescriptor = {|
   sha: string,
   projectId: string,
   branchId: string | "master"
 |};
 
+export type CommitDescriptor = {|
+  sha: string,
+  projectId: string,
+  branchId?: string | "master"
```
bitpshr: Do we need this to be optional? I was thinking `CommitDescriptor` wouldn't include a `branchId` at all. Any endpoint that supports both cases would use a `CommitDescriptor | BranchCommitDescriptor` signature.

tommoor:
If I merge will you take care of new version etc?

PR #176, opened 2019-12-14
fix: Incorrect casing on commits.list parameters

tommoor:
Note: The documentation already listed the camelCased version: https://sdk.abstract.com/docs/abstract-api/#list-all-commits


tommoor:
Will definitely need a patch release once this is merged

PR #184, opened 2019-12-19
feat!: changeset.* return ChangesetResponse

bitpshr:
This pull request updates `changesets.commit` and `changesets.branch` to return a `ChangesetResponse` instead of just a `Changeset`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## [8.0.0-alpha.1](https://github.com/goabstract/abstract-sdk/compare/v8.0.0-alpha.0...v8.0.0-alpha.1) (2019-12-20)
+
+
+### Bug Fixes
+
+* update changeset return value ([567741e](https://github.com/goabstract/abstract-sdk/commit/567741e))
+
+
+
 ## [8.0.0-alpha.0](https://github.com/goabstract/abstract-sdk/compare/v7.0.0...v8.0.0-alpha.0) (2019-12-19)
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,35 +2,6 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
-## [8.0.0-alpha.1](https://github.com/goabstract/abstract-sdk/compare/v8.0.0-alpha.0...v8.0.0-alpha.1) (2019-12-20)
-
-
-### Bug Fixes
-
-* update changeset return value ([567741e](https://github.com/goabstract/abstract-sdk/commit/567741e))
-
-
-
-## [8.0.0-alpha.0](https://github.com/goabstract/abstract-sdk/compare/v7.0.0...v8.0.0-alpha.0) (2019-12-19)
-
-
-### Bug Fixes
-
-* do not bump api version for all endpoints ([f4ce3df](https://github.com/goabstract/abstract-sdk/commit/f4ce3df))
-* move js-sha256 from devDependencies to dependencies ([#178](https://github.com/goabstract/abstract-sdk/issues/178)) ([f9b93b3](https://github.com/goabstract/abstract-sdk/commit/f9b93b3))
-
-
-### Features
-
-* changeset.* return ChangesetResponse ([4d9c3e1](https://github.com/goabstract/abstract-sdk/commit/4d9c3e1))
-
-
-### BREAKING CHANGES
-
-* changeset.* return ChangesetResponse
-
-
-
 ### [7.0.1](https://github.com/goabstract/abstract-sdk/compare/v7.0.0...v7.0.1) (2019-12-16)
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,26 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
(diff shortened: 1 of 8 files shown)


src/types.js
```diff
@@ -860,6 +860,30 @@ export type Changeset = {
...
+export type ChangesetResponse = {
+  changeset: Changeset,
+  colors?: SharedItem[],
+  files: File[],
+  gradients?: SharedItem[],
+  layerStyles?: SharedItem[],
+  layers: Layer[],
+  pages: Page[],
+  sharedData: SharedItem[],
+  textStyles?: SharedItem[]
```
rickharris: This data type is a combination of the v1 and v2 of the changeset response type. I think at this point everything should be using the version that returns `sharedData` and not `colors`, `gradients`, `layerStyles`, and `textStyles`, but we should confirm that for both the api and cli transports.

rickharris: - [v1 in gitapi](https://github.com/goabstract/projects/blob/0b7cc8ad8096746d1e1fa62945a706dba5dae26a/stable/gitapi/web/commit_changeset.go#L138-L148)
- [v2 in gitapi](https://github.com/goabstract/projects/blob/0b7cc8ad8096746d1e1fa62945a706dba5dae26a/stable/gitapi/web/commit_changeset.go#L206-L211)
- [cli](https://github.com/goabstract/projects/blob/0b7cc8ad8096746d1e1fa62945a706dba5dae26a/stable/abstract-cli/cmd/changeset.go#L32-L36) - I don't think we had to maintain a v1 compatibility in cli.

bitpshr: Ah, thanks for this @rickharris! Great information.

tommoor: Looks like we'd need to bump to v19 to get that new format:
https://github.com/goabstract/projects/blob/master/stable/api/config/initializers/versions.rb#L25

tommoor: https://github.com/goabstract/projects/blob/master/stable/api/config/initializers/versions.rb#L25

tommoor: It'll be great to get this covered in the SDK actually, the web app currently reformats the response upon receipt to look like v2 🙈

bitpshr: Updating the API version locally in this case since we're so far behind. When we tackle returning the envelope for all endpoints (pretty high priority, so this will be soon) we can also bump the API version to the latest across the board.

tommoor: Seems like you bumped locally and for all endpoints?

bitpshr: Whoops, fixed. Good catch.

docs/abstract-api.md
```diff
@@ -246,18 +246,30 @@ abstract.branches.info({
 
 A changeset is a group of changes that together form a single, indivisible modification to a project. Changesets include data on all visual and nonvisual changes and provide insight into the differences between two versions of a project.
 
-### The changeset object
+### The ChangesetResponse object
+
+| Property | Type | Description |
+|----------|------|-------------|
+| `changeset` | `Changeset` | Base changeset |
+| `commit` | `Commit` | Commit associated with this changeset |
```
rickharris: This is a weird one... I wonder if it should just be undocumented since it's not included in the cli transport?

src/types.js
```diff
@@ -1490,6 +1511,7 @@ export type Asset = {
   fileFormat: string,
   fileId: string,
   formatName: string,
+
```
rickharris:
```suggestion
```

😂

tommoor:
If you want to publish a prerelease from this branch I'll give it an integration test in desktop

docs/abstract-api.md
bitpshr: Ah, good catch.

src/types.js
bitpshr: _sigh_

bitpshr:
@tommoor `abstract-sdk@8.0.0-alpha.0` is live for testing.

src/endpoints/Changesets.js
```diff
@@ -16,11 +20,12 @@ export default class Changesets extends Endpoint {
...
+    return this.configureRequest<Promise<ChangesetResponse>>(
       {
         api: async () => {
           const response = await this.apiRequest(
-            `projects/${latestDescriptor.projectId}/branches/${latestDescriptor.branchId}/commits/${latestDescriptor.sha}/changeset`
+            `projects/${latestDescriptor.projectId}/branches/${latestDescriptor.branchId}/commits/${latestDescriptor.sha}/changeset`,
+            { headers }
           );
 
           return response.changeset;
```
tommoor: You're still only returning the changeset itself in both these endpoints

tommoor:
I think we should avoid another breaking public version until we've migrated all the neccessary endpoints to return response objects like this btw, we can continue to publish pre-releases in the meantime.

bitpshr:
Good catch, updated @tommoor. Agreed re: prereleases, also.

tommoor:
I think the API is bust, it's sending v19 in the header but still getting the old format back wtf.

davetakahashi-abstract:
Sorry about that all - https://github.com/goabstract/projects/pull/3805 should fix the API.

CHANGELOG.md
```diff
@@ -2,6 +2,26 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## [8.0.0-alpha.0](https://github.com/goabstract/abstract-sdk/compare/v7.0.0...v8.0.0-alpha.0) (2019-12-19)
```
tommoor: Do we need to add notes for alpha releases? Seems like we'd wait and bundle them up into the final release.

bitpshr: This is auto-generated when the `release` NPM script runs. This will be scrapped before landing (same with the alpha version in `package.json`.)

PR #185, opened 2019-12-19
fix: move request options internally

bitpshr:
This pull request is a little housekeeping to consolidate the signature for `Endpoint#configureRequest`. Rather than pass in one argument for endpoint configuration and another for request configuration, everything is passed as one configuration object.

**Note:** Almost all of the changes in this PR are formatting from Prettier, so it's best to hide whitespace changes when viewing it. Flow also strongly types all of the actual code changes here, which should help the review.

### DIFF
--- a/src/endpoints/Endpoint.js
+++ b/src/endpoints/Endpoint.js
@@ -16,8 +16,7 @@ import {
 import type {
   ApiRequestOptions,
   CommandOptions,
-  RequestConfig,
-  RequestOptions
+  RequestConfig
 } from "../types";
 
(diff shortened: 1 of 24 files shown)


src/endpoints/Endpoint.js
```diff
@@ -35,13 +35,11 @@ export default class Endpoint {
     this.options = options;
   }
 
-  configureRequest<T>(
-    config: RequestConfig<T>,
-    requestOptions: RequestOptions = {}
-  ): T {
+  configureRequest<T>(config: RequestConfig<T>): T {
```
bitpshr: 🍖

Issue #186, opened 2019-12-20
Script not running


pascalb09:
Hello,
i have tried to use abstract-sdk in the latest version yesterday version 7.xx, today 8.0.0-alpha.1 and i always get the following error when i call it:
./node_modules/abstract-sdk/dist/paginate.js
Module not found: Can't resolve 'core-js/modules/es7.symbol.async-iterator' in 'C:\dev\project\node_modules\abstract-sdk\dist'.

I tried to call it up with React :)

Do I have to install anything else besides the Abstract SDK?

bitpshr:
Hi @pascalb09, thanks for reaching out. I verified that both `abstract-sdk@7.x.x` and `abstract-sdk@8.0.0-alpha.1` can be installed and that they build properly with a React application generated by `create-react-app@3.3.0`. The SDK installs its own copy of `core-js` as part of its dependency graph.

Could you share a little more information about your application's setup? Specifically, how are you building and running your application, and what does your `package.json` look like?

Thanks in advance for any additional information you can provide.

weyert:
Yes, I am having the same problem. If you want to use this package as part of a Node LTS script it's not working looks like it's not meant to use standalone. You can reproduce it by doing the following:

```
const Abstract = require('abstract-sdk')

const client = new Abstract.Client({
  accessToken: process.env.DEMO_TOKEN,
})
```

weyert:
I have made a private build using this change in the babel config which seems to work for me:

```json
"babel": {
    "presets": [
      "@babel/preset-flow",
      [
        "@babel/preset-env",
        {
          "corejs": 2,
          "useBuiltIns": "usage",
          "targets": {
            "node": 8
          }
        },
        "@babel/preset-react"
      ]
    ],
    "plugins": [
      "@babel/plugin-transform-runtime",
      "@babel/plugin-transform-async-to-generator",
      "@babel/plugin-proposal-object-rest-spread",
      "@babel/plugin-proposal-class-properties"
    ]
  }
```

bitpshr:
Hi @pascalb09 and @weyert, thanks for adding this additional information.

[...]

PR #189, opened 2019-12-27
feat!: remove support for node 8

bitpshr:
This pull request removes support for Node 8 since [its end of life date is 12/31/2019](https://nodejs.org/en/about/releases/). This means that asynchronous iteration support will no longer have to be down-emitted, which will fix the `core-js` issues reported in #186.

References #186


tommoor:
I'm totally fine with this – removed node 8 support from my own projects moons ago.

Issue #193, opened 2019-12-31
Branches endpoint does not return users


tommoor:
Needs moving to the `BranchesResponse` style of response so that all the objects returned by the API can be made available to the SDK consumer. Currently branches and their associated users come down from the API.

Issue #192, opened 2019-12-31
Strategy for policies


tommoor:
Many API endpoints return a `policies` key alongside the `data` key – this is used for knowing the authorization of the current token as it pertains to the objects in the response.

As we continue to use the SDK internally we need access to the policies alongside the data, this could simply be returned as part of the response envelope or treated differently, I'm open to ideas.

amccloud:
It would be nice to avoid exposing the envelope concepts pulically.

If this was cached internally so that subsequent calls to policies (or really any normalized data) can be tracked

```
const params = { projectId: '1' };
const policies = await abstract.policies.show(params); // load policies

// or

const project = await abstract.projects.show(params); // load project and...
const policies = await abstract.policies.show(params); // ...then use cache to access related data
```

tommoor:
@amccloud I think the only issue with this solution, apart from any complexity introduced, is that there are different endpoints that would have identical `params`. Eg, a list of projects might only take the `organizationId` param, but also a single organization query would have the same.

Rather than exposing the envelope itself would you be opposed to the `policies` key being returned by the SDK at the same level as the data, eg it might look like:

```
{
  users: [...],
  branches: [...],
  policies: [...]
}
```

bitpshr:
I think that mixing `policies` with actual data feels a little odd. It may be radical, but I think that we should pass through each API response as-is without any normalization, since this is the same data that consumers would receive if they hit our API directly via HTTP requests (which may happen in the future, and can already be done today.) The idea of using the SDK to hide "internal" data is nice in theory, but it's definitely complicated internal ([and external](https://github.com/goabstract/abstract-sdk/issues/190)) adoption.

amccloud:
@tommoor good point I had that feeling as I typed it though I thought it might not actually be the case. Do we return a list of policies for list? I would expect a org descriptor to always return policies for org to avoid the ambiguity 

I agree with @bitpshr mixing the two and creating yet another type of data here would not be ideal. Maybe we can expose the generic parts of the envelope (no relations) as `Envelope<Project>` which can include policies and other common properties  

Also just throwing this out, what if we provided a `onResponse` callback option to the client?  Could we implement that in ui and normalize all responses straight to redux. You could just use policy selectors as necessary

tommoor:
> Also just throwing this out, what if we provided a onResponse callback option to the client?

[...]

Issue #190, opened 2019-12-30
Files endpoint does not return libraries


naughtandcross:
There is currently no way of discerning which Libraries have been linked to each Project via the API, and the process of checking these connections manually is quite tedious.

Having the ability to obtain the array of Libraries (returning File object references) within the Project object would be super helpful here.

tommoor:
Thanks for the request, this does look like it's an oversight.

tommoor:
@bitpshr this is similar to https://github.com/goabstract/abstract-sdk/pull/184 and is also an internal blocker for similar reasons. We're dropping the `libraries` key of the API response and only returning `files` – we should introduce a `FilesResponse` concept for the files#list endpoint.

bitpshr:
Feature live as of https://github.com/goabstract/abstract-sdk/pull/194.

PR #194, opened 2020-01-03
feat!: expose raw transport responses

bitpshr:
This PR updates all endpoints to expose full responses using a new API.

```js
const branches = await client.branches.list({ ... });
const response = client.unwrap(branches);
```

In the example above, `response` would be the un-sanitized JSON response containing data exactly as it's returned from a given transport. The advantage of using a helper function per @amccloud's suggestion is that consumers aren't forced to grab an arbitrary, potentially-changing property on either the returned object or the returned array, the latter of which felt especially odd.

Internally, we're able to hide the added response key entirely using a non-enumerable object property.

References #192
References #193 
References #190


src/endpoints/Branches.js
```diff
@@ -7,23 +7,28 @@ import type {
...
   info(descriptor: BranchDescriptor, requestOptions: RequestOptions = {}) {
     return this.configureRequest<Promise<Branch>>({
-      api: () => {
-        return this.apiRequest(
+      api: async () => {
+        const response = await this.apiRequest(
           `projects/${descriptor.projectId}/branches/${descriptor.branchId}`
         );
+
+        return wrap(response, response);
```
bitpshr: `branches.info` is a weird method to look at since the `response` is the actual `value`. This was only piped . through the `wrap` flow for consistency. `branches.list` is a better example to check out.

src/response.js
```diff
@@ -0,0 +1,13 @@
+// @flow
+export function wrap(value: any, response: any) {
```
bitpshr: All uses of `wrap` still end up returning strongly-typed values because of upstream generics. We could definitely add a generic here for readability, but it'd force every consumer to redefine the return value type unnecessarily.

src/endpoints/Branches.js
tommoor: Hah, I guess in the case that they are identical we could just pass the single argument and have `wrap` handle that?

src/response.js
```diff
@@ -0,0 +1,13 @@
...
+  value = Array.isArray(value) ? [...value] : { ...response };
```
amccloud: Should this be `{...value}`?

also do we need to clone the value or can we define `_response` on the original value object?

src/response.js
amccloud: Upstream values are `any` so it might not be a poroblem. A generic would allow you to type this locally. We know `value` must be object like for us to define a property on it and `_response` can be `mixed` or also just an object

```js
type Value = {} | [];
type Envelope<T: Value> = T & { _response?: {} };

export function wrap<T: Value>(value: Envelope<T>, response: {}): Envelope<T> {
  Object.defineProperty(value, "_response", {
    enumerable: false,
    value: response
  });

  return value;
}

export function unwrap<T: {}>(value: Envelope<T>) {
  return value._response;
}
```

src/response.js
bitpshr: Great catch re: `{ ...value }`, fixed.

We need to clone the value to prevent a circular reference (adding `_response` to `value` also modifies that data in-place on the original `response`.)

bitpshr: Thinking about this more, we can clone `response` instead, which simplifies this considerably (as well as the typings.)

src/response.js
bitpshr: Good suggestion, updated.

src/endpoints/Branches.js
bitpshr: Updated.

docs/abstract-api.md
```diff
@@ -246,16 +246,6 @@ abstract.branches.info({
...
-### The ChangesetResponse object
-
-| Property | Type | Description |
-|----------|------|-------------|
-| `changeset` | `Changeset` | Base changeset |
-| `files` | `File[]` | Files that changed in this changeset |
-| `layers` | `Layer[]` | Layers that changed in this changeset |
-| `pages` | `Page[]` | Pages that changed in this changeset |
-| `sharedData` | `SharedItem[]` | Shared data that changed in this changeset |
-
```
bitpshr: We can stop maintaining the few `<entity_type>Response` types we had defined now that this data will be exposed via `client.unwrap(entity)`. Admittedly, it does feel weird that we no longer have documentation for these related entity types though. If we keep maintaining the documentation for these response pages (e.g. still maintaining a new concept for each response), why not just maintain and return the type for build-time safety, too?

TLDR; We need to figure out if and how we plan to document this response meta now that the types went away.

bitpshr:
Ready for a pass when you have a chance @tommoor and @amccloud.

docs/abstract-api.md
tommoor: > Admittedly, it does feel weird that we no longer have documentation for these related entity types though

Yep, well I guess this was only supposed to be a stepping stone to move us forward. I could see us having documentation for our actual API also which would help here – but ultimately any objects which can only be accessed through unwrap need their own endpoints, e.g libraries.

bitpshr: Even with separate endpoints for every entity type, we'd still need a way to tell consumers what related entity types are wrapped up in a given response. For example, even though we have a separate endpoint for `files` already, there's no indication that `files` even exists in the wrapped response for `changesets`. Internal caching may help, but I'm still a little skeptical of a cache-based approach (e.g. to get pages related to changesets, a consumer would have to fetch all changesets, then use the same `BranchDescriptor` to fetch all files, then iterate over each file and request each page, etc.)

I agree that we're getting to the point where we need API documentation to better support this need and also external contributions.

tommoor:
Wanna publish a prerelease and I'll try and integrate internally?

bitpshr:
Definitely @tommoor. `abstract-sdk@8.0.0-beta.3` is now live for testing.

Issue #182, opened 2019-12-19
API Transport files.raw endpoint


MichaelZaporozhets:
Is there any reason the `files.raw` endpoint is not available for the http api?

bitpshr:
Hi @MichaelZaporozhets, thanks for reaching out. To date, file export functionality only existed in Abstract's CLI and didn't exist in the API at all. We're actively working on (and almost done!) adding this functionality to the API, and we hope to land this endpoint in the SDK within the next few weeks.

MichaelZaporozhets:
> Hi @MichaelZaporozhets, thanks for reaching out. To date, file export functionality only existed in Abstract's CLI and didn't exist in the API at all. We're actively working on (and almost done!) adding this functionality to the API, and we hope to land this endpoint in the SDK within the next few weeks.

Ah- I see! Glad to hear it's being worked on- i'd be very willing to help out once it's ready in API 🙂

It might be outside the scope of this repository, but a separate API doc might be useful to enable more community engagement.

weyert:
Happy to hear that! Looking forward to this feature :D

tommoor:
> a separate API doc might be useful to enable more community engagement.

Could you say more about what you mean here?

MichaelZaporozhets:
@tommoor Given that the web api development track is internal and is/can be developed asynchronously to the sdk- creating a spec portal for the web api would allow external developers to develop/contribute to the sdk against that spec/contract, rather than leaving pressure of feature sync on the team that maintains the sdk and/or api. 

Lots of assumptions here and slashes to match but I'm hoping it's clear 🙂

weyert:
I would love to know what the timings are to support downloading Sketch files from Abstract via the API? As it's hard to convince people to run a Mac server just to be able to generate design tokens from a Sketch file via our CI environment.

bitpshr:
Hi @weyert. This functionality will be part of the upcoming `abstract-sdk@8.0.0` major release, which should be live by the end of the week. We'll be sure to comment on this issue when that happens to let you know when it's available for use.

bitpshr:
API file export is now live in the latest `abstract-sdk@8.0.0-beta.7` prerelease and will be included in the upcoming `abstract-sdk@8.0.0` major release.

PR #199, opened 2020-01-07
feat!: api file export

bitpshr:
This pull request adds file export functionality to the API transport.

References #182


src/endpoints/Files.js
```diff
@@ -81,7 +85,69 @@ export default class Files extends Endpoint {
...
+            return arrayBuffer;
+          } else if (
+            callCount * EXPORT_TIMEOUT >= MAX_EXPORT_DURATION ||
+            exportJob.status === "failed"
+          ) {
+            throw new FileExportError(file.id, exportJob.id);
+          } else {
+            callCount++;
+            await new Promise(resolve => setTimeout(resolve, EXPORT_TIMEOUT));
+            return checkStatus();
```
MichaelZaporozhets: rather than mutating `callCount` on iterations, why not something like:
```
const checkStatus = async (callCount) => {
...
const currentCount = callCount ? callCount : 0
const nextCount = callCount + 1;
...
return checkStatus(nextCount);
```

bitpshr: Thanks for the feedback @MichaelZaporozhets, good suggestion. Updated.

bitpshr:
Rebased and ready for a pass whenever there's time. Will be great to get this into a prerelease for some additional battle-testing.

tommoor:
@bitpshr were we waiting on additional protections on the API side?

bitpshr:
The API is [already throttling](https://github.com/goabstract/projects/blob/master/stable/api/app/helpers/sdk_request_filter.rb#L53) export requests, so I think we should be OK.

src/endpoints/Files.js
```diff
@@ -80,7 +83,67 @@ export default class Files extends Endpoint {
       descriptor
     );
 
-    return this.configureRequest<Promise<void>>({
+    return this.configureRequest<Promise<ArrayBuffer | void>>({
```
tommoor: We should update the documentation also to show that this method is available through API. Currently that section is called "Retrieve a Sketch file", considering we now support XD should probably rename that.

src/endpoints/Files.js
```diff
@@ -80,7 +83,67 @@ export default class Files extends Endpoint {
...
+              method: "POST",
+              body: { ...latestDescriptor, export_id: exportId }
+            }
+          );
+        };
+
+        const file = await this.info(latestDescriptor);
+        let exportJob = await exportRequest();
+
+        const checkStatus = async (count: number) => {
```
tommoor: I'm concerned that between the export and the download of potentially hundreds of Mb's this will take quite some time, could we provide a callback in the options to report at least progress between "exporting" | "downloading" ?

bitpshr: Just to clarify, when you say "download", you mean writing to disk, right? If so, we could add insight into this process, but we don't offer any such insight in the CLI transport, which complicates the typings of the request options object and adds odd API asymmetry. Once we have the array buffer (which we get as soon as the server export is done), writing to disk (even for large files) took much less time than the export itself, FWIW.

bitpshr: Just to clarify, when you say "downloading",  you mean the time between when the export is complete server-side and when the response to get the array buffer completes? We could expose an options callback API to track this as requested, but we don't offer any similar insight with the CLI transport, which makes for a weirdly asymmetrical API.

How do we track progress of the CLI operation in client applications today?

tommoor: As I understand it there are three distinct phases for the API transport:

1. Compiling the file on the server
1. Downloading the file to the client
1. Writing the file to disk

For the CLI transport, I guess it essentially skips the second step and the other two have been fast enough that we didn't worry about progress. I don't think this should necessarily block this PR, but 
if we end up using this for the partial sync stuff we'll definitely need a way to display progress.

bitpshr: Thanks for this feedback. `files.raw` now accepts `RawProgressOptions` instead of just `RawOptions`, which supports an `onProgress(receivedBytes: number, totalBytes: number)` callback. This gives insight into the underlying download operation. This _should_ be sufficient for most use cases, since neither the actual file generation nor the disk writing take significant time, and since progress insight isn't exposed into these steps in the existing CLI.

See an example [here](https://github.com/goabstract/abstract-sdk/pull/199/files#diff-faca079f15c779be0f73b337eec64767R834-R846).

bitpshr:
Thanks for the review so far @tommoor. An `onProgress` option was added and documentation updated. Ready for another pass.

bitpshr:
Any thoughts on this type of progress exposure @tommoor? No major rush here, but this PR blocks a few internal and external initiatives. Always happy to hop on a call since it's a bit of a dense PR.

src/endpoints/Files.js
tommoor: Right on – happy with this for a first phase, as discussed we'll probably want to add the additional progress events for file generation later both on API and CLI but seeing as neither of these exist today we'll go with what we have!

docs/abstract-api.md
```diff
@@ -794,11 +794,11 @@ abstract.files.info({
...
-### Retrieve a Sketch file
+### Export a file
 
-![CLI][cli-icon]
+![CLI][cli-icon] ![API][api-icon]
 
-`files.raw(FileDescriptor, RawOptions): Promise<ArrayBuffer>`
+`files.raw(FileDescriptor, RawProgressOptions): Promise<ArrayBuffer | void>`
 
 Retrieve a Sketch file from Abstract based on its file ID and save it to disk. Files will be saved to the current working directory by default, but a custom `filename` option can be used to customize this location.
```
tommoor: Not directly edited on this branch but this doesn't seem right – or at least we need to mention that `disableWrite` is required too?

> but a custom `filename` option can be used to customize this location.

src/endpoints/Endpoint.js
```diff
@@ -19,6 +20,7 @@ import type {
   RequestConfig
 } from "../types";
 
+const cliPath = require("@elasticprojects/abstract-cli");
```
tommoor: Was this not within the cliRequest block to prevent issues in browsers?

src/endpoints/Files.js
```diff
@@ -67,20 +70,81 @@ export default class Files extends Endpoint {
...
+
+            /* istanbul ignore if */
+            if (isNodeEnvironment() && !disableWrite) {
+              const diskLocation = filename || `${file.name}.sketch`;
+              fs.writeFile(diskLocation, Buffer.from(arrayBuffer));
+            }
+
+            return arrayBuffer;
+          } else if (
+            count * EXPORT_TIMEOUT >= MAX_EXPORT_DURATION ||
```
tommoor: The naming of `EXPORT_TIMEOUT` threw me for a minute while reviewing this, better as something along the lines of `EXPORT_STATUS_CHECK_INTERVAL` ?

src/endpoints/Files.js
```diff
@@ -67,20 +70,81 @@ export default class Files extends Endpoint {
...
+              ""
+            );
+
+            const arrayBuffer = await this.apiRequest(
+              filePath,
+              {
+                headers: {
+                  Accept: undefined,
+                  "Content-Type": undefined,
+                  "Abstract-Api-Version": undefined
```
tommoor: Why do we explicitly unset the API version for this endpoint?

docs/abstract-api.md
bitpshr: `disableWrite` is meant to be used when you want to do something special with thee `ArrayBuffer`, so the `filename` won't do anything in this case.

src/endpoints/Files.js
bitpshr: Leftover from another endpoint, removed.

src/endpoints/Endpoint.js
bitpshr: This should be OK since we're handling this in `package.json` like other Node-specific modules using the `browser` property.

Issue #180, opened 2019-12-17
Full Project CRUD?


benmcginnis:
Hi folks, I'm looking to automate some manual work that goes into new projects for our team and a lot of it is going to depend on being able to create / update Abstract projects via your API.  Any ETA on when you could add a create endpoint here? Would a pull request be appreciated?

https://github.com/goabstract/abstract-sdk/blob/a21c7af3f75d459ef196fe6729b1a7c0d63a441a/src/endpoints/Projects.js#L16

tommoor:
Hey @benmcginnis – totally understand the need for this. Unfortunately a PR here won't do much good as the issue is a lack of support on our API for creating projects right now, there is a dependency on the desktop client that we need to resolve first.

benmcginnis:
Thanks for letting me know @tommoor.  Any thoughts on priority?  I'm specing out a full workflow through QA automation that would depend on this to work.

tommoor:
We're working on enabling project creation on the web which will be dependent on the same endpoint – I'm not directly involved in the work so can't give an estimate, other than to say the project is ongoing.

benmcginnis:
Thank you!

bitpshr:
Thanks for helping us prioritize this feature internally @benmcginnis. Don't hesitate to reach out with any additional questions or feature requests.

benmcginnis:
Thanks, happy to help!

MichaelZaporozhets:
this would be awesome. +1

bitpshr:
Addressed as part of #201.

PR #201, opened 2020-01-07
feat: support full project CRUD

bitpshr:
This pull request updates the API transport by adding support for project creation, updating, deletion, archiving, and unarchiving.

Requires https://github.com/goabstract/projects/pull/3841
References #180

### DIFF
--- a/abstract-sdk.d.ts
+++ b/abstract-sdk.d.ts
@@ -296,6 +296,28 @@ interface Projects extends Endpoint {
       sectionId?: string
     }
   ): Promise<Project[]>;
+  create(
+    descriptor: OrganizationDescriptor,
+    project: NewProject,
+    requestOptions?: RequestOptions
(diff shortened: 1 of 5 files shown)


tommoor:
Unfortunately we can't enable project creation via the SDK until the underlying repository is setup entirely server-side, currently the API endpoint relies on the fact that the desktop client will create the repository with a `PROJECT_CREATED` commit and push that to the server.

I think we'll need a new API version that enables gitnode to perform this operation, rather than the client, and then use that API version here before it's mergeable.

bitpshr:
Ah, thanks for this @tommoor. I will try to get the gitnode stuff in place.

tommoor:
Removing review request until backend is ready.

bitpshr:
Ready for another pass now that the backend changes are live and the API version has been bumped @tommoor. Thanks in advance!

PR #203, opened 2020-01-14
fix: Ensure policies are returned in branches payload

tommoor:
fix: Commits list api endpoint (broken in v8 prerelease)
fix: Ensure policies are returned in unwrapped branches payload


src/endpoints/Commits.js
```diff
@@ -27,10 +27,10 @@ export default class Commits extends Endpoint {
         // loading commits with a share token requires a branchId so this
         // route is maintained for that circumstance
         const response = descriptor.branchId
-          ? this.apiRequest(
```
tommoor: Found this bug whilst adjusting the tests

Issue #202, opened 2020-01-11
Branches endpoint should allow search without projectId


tommoor:
The project descriptor should be optional if a `search` parameter is present, we use this to search for branches across all projects in the quick jump internally.

amccloud:
That's an interesting case and seems like a solid way forward if you search projects across organizations

PR #204, opened 2020-01-15
fix: Allow search param for branches

tommoor:
@bitpshr I know you're assigned but I needed to move this forward faster – is this approach dumb? Feel free to completely re-do. Mainly I wasn't sure how to deal with the API-only parameters.

Related #202

### DIFF
--- a/src/endpoints/Branches.js
+++ b/src/endpoints/Branches.js
@@ -42,32 +42,43 @@ export default class Branches extends Endpoint {
   }
 
   list(
-    descriptor: ProjectDescriptor,
+    descriptor: ?ProjectDescriptor,
     options: {
       ...RequestOptions,
-      filter?: "active" | "archived" | "mine"
+      filter?: "active" | "archived" | "mine",
+      search?: string
     } = {}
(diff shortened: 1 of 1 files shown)


src/endpoints/Branches.js
```diff
@@ -42,32 +42,43 @@ export default class Branches extends Endpoint {
...
       },
 
       cli: async () => {
+        const projectId = descriptor ? descriptor.projectId : undefined;
+        if (!projectId) {
+          throw new Error("projectId required with CLI transport");
+        }
+        if (search) {
+          throw new Error("search not supported with CLI transport");
+        }
```
bitpshr: I would simplify this a bit: throw an error about a `ProjectDescriptor` being required for the CLI transport if a descriptor is not passed. I wouldn't check for `search` (or do the `projectId` ternary) at all.

tommoor:
@bitpshr would you? I'm stacked today.

bitpshr:
Superseded by https://github.com/goabstract/abstract-sdk/pull/206.

PR #206, opened 2020-01-15
feat: support branch searching

bitpshr:
This pull request updates the `branches.list` endpoint to support name-based searches when using the API transport.

References #202 
Requires https://github.com/goabstract/projects/pull/3891


bitpshr:
@tommoor: This requires https://github.com/goabstract/projects/pull/3891 if you have a chance to check that out as well.

PR #207, opened 2020-01-15
feat: add layer dependency filtering options

rickharris:
Depends on https://github.com/goabstract/projects/pull/3894, posting the same explanation here for visibility

This adds two new filtering options for layers:

- `excludeLibraryDependencies` - this will make it so that symbols from external libraries are not included in the list of layers. These are the items that we show as "dependencies" in the file list and aren't shown by default unless you click on them.
- `onlyLibraryDependencies` - the opposite of `excludeLibraryDependencies`. You can choose this option if you need to load the library dependencies separately.

These options should help our pagination of layers be a little more intuitive so that we don't load pages of data that aren't even shown in the screen you're looking at.

### DIFF
--- a/docs/abstract-api.md
+++ b/docs/abstract-api.md
@@ -851,7 +851,7 @@ A layer is a container for designs. In Sketch a layer usually represents an artb
 
 ![CLI][cli-icon] ![API][api-icon]
 
-`layers.list(FileDescriptor | PageDescriptor, ListOptions): Promise<Layer[]>`
+`layers.list(FileDescriptor | PageDescriptor, LayersListOptions): Promise<Layer[]>`
 
 List the layers for a file at a commit
 
@@ -875,7 +875,9 @@ abstract.layers.list({
   sha: "fb7e9b50da6c330fc43ffb369616f0cd1fa92cc2"
 }, {
   limit: 25,
-  offset: 0
+  offset: 0,
+  excludeLibraryDependencies: false,
+  onlyLibraryDependencies: false
 });
 ```
 
--- a/docs/abstract-api.md
+++ b/docs/abstract-api.md
@@ -875,9 +875,23 @@ abstract.layers.list({
   sha: "fb7e9b50da6c330fc43ffb369616f0cd1fa92cc2"
 }, {
   limit: 25,
+  offset: 0
+});
+```
+
+A file can also contain layers from external libraries. If you'd like to only see the layers in this file and not the external library elements they depend upon, use the `fromLibraries` option…
+
+```js
+abstract.layers.list({
+  projectId: "616daa90-1736-11e8-b8b0-8d1fec7aef78",
+  branchId: "master",
+  fileId: "51DE7CD1-ECDC-473C-B30E-62AE913743B7",
+  pageId: "7D2D2599-9B3F-49BC-9F86-9D9D532F143A",
+  sha: "fb7e9b50da6c330fc43ffb369616f0cd1fa92cc2"
+}, {
+  limit: 25,
   offset: 0,
-  excludeLibraryDependencies: false,
-  onlyLibraryDependencies: false
+  fromLibraries: "exclude"
 });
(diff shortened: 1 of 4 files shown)


rickharris:
I can't say I understand this test failure. @bitpshr have you seen that `MultiError` before?

src/endpoints/Layers.js
```diff
@@ -48,19 +54,26 @@ export default class Layers extends Endpoint {
...
+      ...requestOptions
+    } = options;
     const latestDescriptor = await this.client.descriptors.getLatestDescriptor(
       descriptor
     );
 
     return this.configureRequest<Promise<Layer[]>>({
       api: async () => {
         const query = querystring.stringify({
-          ...latestDescriptor,
```
rickharris: I removed this because I noticed it while using this endpoint, that the parameters built into the URL were also being passed as query parameters to the API. It's also currently built into the tests as mocks. @bitpshr is this something we're comfortable removing or is it here for a reason?

docs/abstract-api.md
```diff
@@ -851,7 +851,7 @@ A layer is a container for designs. In Sketch a layer usually represents an artb
 
 ![CLI][cli-icon] ![API][api-icon]
 
-`layers.list(FileDescriptor | PageDescriptor, ListOptions): Promise<Layer[]>`
+`layers.list(FileDescriptor | PageDescriptor, LayersListOptions): Promise<Layer[]>`
```
rickharris: There's no other documentation for what this type is, should I add it somewhere?

docs/abstract-api.md
```diff
@@ -875,7 +875,9 @@ abstract.layers.list({
   sha: "fb7e9b50da6c330fc43ffb369616f0cd1fa92cc2"
 }, {
   limit: 25,
-  offset: 0
+  offset: 0,
+  excludeLibraryDependencies: false,
+  onlyLibraryDependencies: false
```
rickharris: I added these here just to say that they exist but they're probably very out of place.

src/endpoints/Layers.js
bitpshr: If we remove this, I think we need to explicitly pass `sha` (and potentially `pageId`) into the query string.

docs/abstract-api.md
bitpshr: Our documentation is a mess, this will hopefully be fixed soon. For now, feel free to define this where `ListOptions` is defined towards the end of this document.

docs/abstract-api.md
bitpshr: This is fine for now. When we generate API documentation (planned), this will be shown in the right spot.

tommoor:
Did you consider using a single `filter` parameter for this? It feels like it would be more suited considering it doesn't make sense to set both flags at once… eg:

- `filter=undefined` = returns all results as today
- `filter=all` = returns all results as today
- `filter=dependencies` = returns only dependencies
- `filter=normal` = returns regular non-dependencies ("normal" is up for grabs here)

rickharris:
The same was suggested in the backend PR https://github.com/goabstract/projects/pull/3894#discussion_r367148849

I'll update

tommoor:
Oh, lol – that was later down my list of PR's to get to 😄

rickharris:
updated the option to `fromLibrary: "include" | "exclude" | "only"` in 2ecbc1ce587dbe4280f7dde75bf644ad8e367fe8

src/endpoints/Layers.js
```diff
@@ -11,6 +11,11 @@ import type {
 import Endpoint from "../endpoints/Endpoint";
 import { wrap } from "../util/helpers";
 
+type LayersListOptions = {
+  ...ListOptions,
+  fromLibraries?: "include" | "excluce" | "only"
```
tommoor:
```suggestion
  fromLibraries?: "include" | "exclude" | "only"
```

PR #208, opened 2020-01-21
fix: pass branch id to cli for data endpoint

rickharris:
When using the cli transport for layer data, the following error is getting bubbled up from `abstract-sdk`:

> Error runtime error: index out of range [4] with length 4 ### CLI command: abstract-cli layer data [projectId] [sha] [fileId] [layerId]

We aren't currently passing in a branch id, even though the version of abstract-cli we depend upon uses it.

### DIFF
--- a/src/endpoints/Data.js
+++ b/src/endpoints/Data.js
@@ -29,6 +29,7 @@ export default class Data extends Endpoint {
           "layer",
           "data",
           latestDescriptor.projectId,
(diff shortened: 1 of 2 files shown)


bitpshr:
Looks good. Let me know if you'd like me to get this into the latest prerelease, or feel free to do it yourself if you prefer:

```
yarn release --prerelease beta --skip.changelog && git push origin master --follow-tags
```

PR #209, opened 2020-01-23
feat: add layer filtering options to cli transport

rickharris:
Depends on https://github.com/goabstract/projects/pull/3929 and a new version of abstract-cli being released

This PR adds filtering options to `layers.list` for the `cli` transport so that that same filtering options are available as the `api` transport. Previously, sdk would allow you to pass filtering options but would ignore them for the `cli` transport.

Documentation should already be up-to-date since we already support these options for the `api` transport.

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
   },
   "dependencies": {
     "@babel/runtime-corejs2": "^7.3.4",
-    "@elasticprojects/abstract-cli": "^2.3.0",
+    "@elasticprojects/abstract-cli": "^3.5.0",
     "cross-fetch": "^3.0.1",
     "debug": "^4.0.1",
     "flow-typed": "^2.5.1",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "abstract-sdk",
-  "version": "8.0.0-beta.7",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "abstract-sdk",
-  "version": "8.0.0-beta.7",
+  "version": "8.0.0-beta.9",
```
rickharris: Just included this so we'd have the right package.json version on master.

package.json
```diff
@@ -65,7 +65,7 @@
   },
   "dependencies": {
     "@babel/runtime-corejs2": "^7.3.4",
-    "@elasticprojects/abstract-cli": "^2.3.0",
+    "@elasticprojects/abstract-cli": "^3.5.0",
```
rickharris: If we trust our [cli release notes](https://github.com/goabstract/projects/blob/master/stable/abstract-cli/CHANGELOG.md#300-2019-11-18), there shouldn't be any compatibility issues

tommoor: Because the breaking change is in the CLI projects endpoint and that's not used by the SDK?

rickharris: yep yep

Issue #212, opened 2020-01-31
SDK embed generator not working


mkeys:
The embed generator on the [SDK site](https://sdk.abstract.com/docs/embed/) isn't working. I get the error "Invalid share link. Please try again..."

Here is the PSL link I tried: https://share.goabstract.com/aaec8bba-f473-4f64-96e7-bff41c70ff8a?mode=design&sha=576401cfbcc90bdde3a2a16500c1f4e74dd02c80

tommoor:
Thanks @mkeys !

Sidenote: It'd be nice if we ensured the generated embeds have the new domain

bitpshr:
The embed generator at https://developer.abstract.com/docs/embed/#generating-an-embed should now handle the above URL properly. Thanks for filing this issue @mkeys!

PR #213, opened 2020-02-03
fix: embed generator

bitpshr:
This pull request updates the regular expression logic used by the embed code generator to be less brittle and to properly handle valid URLs with valid parameters.

References #212

### DIFF
--- a/docs/embed.md
+++ b/docs/embed.md
@@ -20,16 +20,16 @@ Embeds can be used to display interactive, up-to-date previews for layers and co
       const copy = document.querySelector('#copy');
       const input = document.querySelector('#embed-input');
       const waiting = document.querySelector('#waiting');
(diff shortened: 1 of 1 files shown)


docs/embed.md
```diff
@@ -20,16 +20,16 @@ Embeds can be used to display interactive, up-to-date previews for layers and co
       const copy = document.querySelector('#copy');
       const input = document.querySelector('#embed-input');
       const waiting = document.querySelector('#waiting');
+      const shareLinkPattern = /^https?:\/\/share.(go)?abstract.com/;
```
tommoor:
```suggestion
      const shareLinkPattern = /^https?:\/\/share\.(go)?abstract\.com/;
```

Need to escape the periods otherwise it's being more permissive than you'd expect

PR #215, opened 2020-02-11
refactor: stop using deprecated CLI commands

Nivl:
Fairly straightforward PR. We deprecated a certain amount of commands in the CLI those last 6 to 12 months, this PR makes sure we're no longer using them.

### DIFF
--- a/src/endpoints/Branches.js
+++ b/src/endpoints/Branches.js
@@ -30,10 +30,10 @@ export default class Branches extends Endpoint {
 
       cli: async () => {
         const response = await this.cliRequest([
-          "branch",
-          "load",
-          descriptor.projectId,
-          descriptor.branchId
+          "branches",
+          "get",
+          descriptor.branchId,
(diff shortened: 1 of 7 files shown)


src/endpoints/Branches.js
```diff
@@ -30,10 +30,10 @@ export default class Branches extends Endpoint {
...
       cli: async () => {
         const response = await this.cliRequest([
-          "branch",
-          "load",
-          descriptor.projectId,
-          descriptor.branchId
+          "branches",
+          "get",
+          descriptor.branchId,
+          ...["--project-id", descriptor.projectId]
```
bitpshr: Since we know that a project ID will always be passed, we don't need to spread an array. Instead, we can just list `"--project-id", descriptor.projectId`. This comment applies throughout and is definitely nonblocking.

```suggestion
          "--project-id",
          descriptor.projectId
```

PR #220, opened 2020-02-25
chore: add custom api version to collections request

timc1:
This PR adds the ability to use a custom api version on the collections request. 

Part of https://github.com/goabstract/projects/pull/4159

Project reference: https://github.com/goabstract/ui/pull/2451

### DIFF
--- a/src/types.js
+++ b/src/types.js
@@ -101,7 +101,7 @@ export type ApiRequestOptions = {
 
 export type RequestOptions = {
   transportMode?: ("api" | "cli")[],
-  headers?: { [string]: string }
+  _version?: number
 };
 
 export type RequestConfig<T> = {
--- a/src/types.js
+++ b/src/types.js
@@ -100,7 +100,8 @@ export type ApiRequestOptions = {
 };
 
 export type RequestOptions = {
-  transportMode?: ("api" | "cli")[]
+  transportMode?: ("api" | "cli")[],
+  headers?: { [string]: string }
 };
 
 export type RequestConfig<T> = {
--- a/src/endpoints/Collections.js
+++ b/src/endpoints/Collections.js
@@ -14,9 +14,7 @@ import Endpoint from "../endpoints/Endpoint";
 import { wrap } from "../util/helpers";
 
 // Version 16 returns cached thumbnails
-const headers = {
-  "Abstract-Api-Version": "16"
-};
+const API_VERSION = 16;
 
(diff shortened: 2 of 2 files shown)


src/types.js
```diff
@@ -100,7 +100,8 @@ export type ApiRequestOptions = {
 };
 
 export type RequestOptions = {
-  transportMode?: ("api" | "cli")[]
+  transportMode?: ("api" | "cli")[],
+  headers?: { [string]: string }
```
rickharris: do we want to introduce HTTP headers into a system that can request data from an HTTP API or a non-HTTP cli? I would think we'd want to hard-code the the API version per-endpoint (confirmed with @timc1 that the intention is to use this to set the API version) so that we can return proper flow-types and all that rather than allowing SDK consumers to specify API versions.

cc @amccloud @bitpshr

tommoor: > I would think we'd want to hard-code the the API version per-endpoint

Agreed. Although it does raise the question of the best way to achieve that when the old version is still live behind a feature flag. I could see a more specific param for that request making sense, eg `includeTextType` which internally is mapped to an API version?

bitpshr: I agree with @rickharris that we should continue abstracting the idea of underlying API versions from consumers. I think we should either use a specific API option for the case where old and new APIs are concurrently live for the same endpoint, or just hardcode the needed API version like we do elsewere.

amccloud: @rickharris I agree about the transport agnostic options for the client. We wouldn't want to hardcode though, as we'll still want to experiment without forcing the new version as the default

@bitpshr I like that second option. Version number is how we reason about these changes elsewhere so seems ideal to keep it the same. @tommoor would this work?

```suggestion
  version?:  number
```

tommoor: It would, but I'd still prefer to keep version number concerns internal to the SDK and instead expose it as a named option relevant to the feature.

Whatever crowd consensus is though.

amccloud: @tommoor that sounds nice but ultimately might not be how we want to end up managing experimental features in the sdk. If you had a bunch of flags that can be combined dynamically in adhoc ways, it would be hard to provide the correct type for that endpoint

My vote would be to add a `version` property (or `_version`) but not document it as public API. The public interface features is controlled by the sdk's semver. We could introduce things to the sdk as the become ready, typed, documented, and out of the box without the need to set flags, verisons, etc.

PR #221, opened 2020-02-25
fix: support branch pagination

### LINKED ISSUE #219: branches.list missing pagination
https://github.com/goabstract/abstract-sdk/blob/master/src/endpoints/Branches.js#L57-L60

We're not passing limit/offset through to the branches list endpoints. Seems to be the likely cause of this:
https://goabstract.atlassian.net/browse/PLATFORM-1051

Not sure how this was missed when I integrated with the main app 🙈

bitpshr:
This pull request updates the branches endpoint to properly pass pagination query parameters.

Fixes #219

### DIFF
--- a/src/endpoints/Branches.js
+++ b/src/endpoints/Branches.js
@@ -3,6 +3,7 @@ import querystring from "query-string";
 import type {
   Branch,
   BranchDescriptor,
+  ListOptions,
   ProjectDescriptor,
   RequestOptions
 } from "../types";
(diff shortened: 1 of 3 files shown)


bitpshr:
Note: `abstract-sdk@8.0.0-beta.12` is live for testing with this change.

src/endpoints/Branches.js
```diff
@@ -45,16 +46,16 @@ export default class Branches extends Endpoint {
...
       search?: string
     } = {}
   ) {
-    const { filter, search, ...requestOptions } = options;
+    const { limit, offset, filter, search, ...requestOptions } = options;
 
     return this.configureRequest<Promise<Branch[]>>({
       api: async () => {
-        const query = querystring.stringify({ filter, search });
+        const query = querystring.stringify({ limit, offset, filter, search });
```
tommoor: We could really do with the same params for CLI, I'll file a separate issue.

bitpshr: Ah, didn't realize the CLI supported pagination. On it!

tommoor: Sorry, it doesn't right now – that was ambiguous. I filed an internal issue to add them to CLI

PR #226, opened 2020-03-06
Bugfix/ie11 support

berezovskycom:


berezovskycom:
Hey @amccloud! It looks like we haven't used sdk version of this pr in ui/web repo, hence our Abstract website would not support old versions of browsers, because ui repo uses sdk master branch.
I compiled sdk using webpack, tested it on ie11, and everything seems to work fine.
Let's push changes into master and see if it works

PR #227, opened 2020-03-08
fix: Type ErrorData in abstract-sdk.d.ts

mascardoso:
#228


mascardoso:
@amccloud thank you. A quick one. Not sure when would this be available? How does your release process work? Thank you.

tommoor:
Hi, the fix was released as part of `8.0.0-beta.17`

PR #231, opened 2020-03-11
Update usage.md

vjandrei:
Can it be changed to be not es6 import?


berezovskycom:
Hi @vjandrei!
We tend to use the latest JavaScript principles, and importing with a require() seems like not the case.

PR #232, opened 2020-03-11
feat: branches merge state

johncoder:
Adds the ability to query the branch merge state for both the API and CLI transports.

Note that CLI only provides the `state` while the API returns more detailed information.

### DIFF
--- a/abstract-sdk.d.ts
+++ b/abstract-sdk.d.ts
@@ -153,7 +153,7 @@ interface Branches extends Endpoint {
     descriptor?: ProjectDescriptor,
     options?: RequestOptions & { filter?: "active" | "archived" | "mine", search?: string }
   ): Promise<Branch[]>;
-  mergeState(descriptor: BranchDescriptor, parent?: string): Promise<BranchMergeState>;
+  mergeState(descriptor: BranchDescriptor, options?: { parent?: string }): Promise<BranchMergeState>;
 }
 
 interface Changesets extends Endpoint {
--- a/abstract-sdk.d.ts
+++ b/abstract-sdk.d.ts
@@ -153,6 +153,7 @@ interface Branches extends Endpoint {
     descriptor?: ProjectDescriptor,
     options?: RequestOptions & { filter?: "active" | "archived" | "mine", search?: string }
   ): Promise<Branch[]>;
+  mergeState(descriptor: BranchDescriptor, parent?: string): Promise<BranchMergeState>;
 }
 
 interface Changesets extends Endpoint {
--- a/docs/abstract-api.md
+++ b/docs/abstract-api.md
@@ -271,7 +271,7 @@ A branch merge state is a description of whether a branch can be cleanly merged 
 
 ![CLI][cli-icon] ![API][api-icon]
 
-`branches.mergeState(BranchDescriptor, string): Promise<BranchMergeState>`
+`branches.mergeState(BranchDescriptor, options?: { parent?: string }): Promise<BranchMergeState>`
 
 Load the merge state for a specific branch in a project.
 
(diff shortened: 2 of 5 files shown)


src/endpoints/Branches.js
```diff
@@ -81,4 +82,29 @@ export default class Branches extends Endpoint {
       requestOptions
     });
   }
+
+  mergeState(descriptor: BranchDescriptor, parent?: string) {
+    return this.configureRequest<Promise<BranchMergeState>>({
+      api: async () => {
+        let requestUrl = `projects/${descriptor.projectId}/branches/${descriptor.branchId}/merge_state`;
+        if (parent) {
+          const query = querystring.stringify({ parent_id: parent });
```
tommoor: This query parameter should have been camelCase on the API side to match elsewhere… fix up before it reaches prod?

johncoder: sure thing

johncoder: API has been updated.

src/types.js
```diff
@@ -785,6 +785,18 @@ export type Branch = {
...
+export type MergeState = "CLEAN" | "NEEDS_UPDATE" | "NEEDS_REMOTE_UPDATE";
+
+export type BranchMergeState = {
+  state: MergeState,
+  parentId?: string,
+  parentCommit?: string,
+  branchId?: string,
+  branchCommit?: string,
+  ahead?: number,
+  behind?: number
```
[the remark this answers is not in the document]
tommoor: Unfortunately, but seems like we could update cli in the future to return these too

src/endpoints/Branches.js
```diff
@@ -81,4 +82,29 @@ export default class Branches extends Endpoint {
       requestOptions
     });
   }
+
+  mergeState(descriptor: BranchDescriptor, parent?: string) {
```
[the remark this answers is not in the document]
tommoor: a-gree

src/types.js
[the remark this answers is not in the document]
johncoder: Yeah, it'll all be provided for the API transport. It ought to be helpful for understanding what commits were used to determine the merge state.

src/endpoints/Branches.js
```diff
@@ -81,4 +82,29 @@ export default class Branches extends Endpoint {
...
+        let requestUrl = `projects/${descriptor.projectId}/branches/${descriptor.branchId}/merge_state`;
+        if (parent) {
+          const query = querystring.stringify({ parentId: parent });
+          requestUrl = `${requestUrl}?${query}`;
+        }
+        const response = await this.apiRequest(requestUrl, { headers });
+        return wrap(response.data, response);
+      },
+
+      cli: async () => {
```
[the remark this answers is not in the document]
johncoder: Not yet. Anything specific I should do on this point? I can at least include a comment explaining what differs between the two transports, if you think that'd help.

abstract-sdk.d.ts
```diff
@@ -153,6 +153,7 @@ interface Branches extends Endpoint {
     descriptor?: ProjectDescriptor,
     options?: RequestOptions & { filter?: "active" | "archived" | "mine", search?: string }
   ): Promise<Branch[]>;
+  mergeState(descriptor: BranchDescriptor, parent?: string): Promise<BranchMergeState>;
```
amccloud: 👀 outdated

docs/abstract-api.md
```diff
@@ -251,6 +251,38 @@ abstract.branches.info({
...
+Load the merge state for a specific branch in a project.
+
+```js
+abstract.branches.mergeState({
+  projectId: "616daa90-1736-11e8-b8b0-8d1fec7aef78",
+  branchId: "252e9141-5cfc-4f4e-823b-a85f4cfd9a32"
+});
+```
+
+> Note: the optional fields in `BranchMergeState` are only available when using the [API transport](/docs/transports). The API transport only returns states `CLEAN` or `NEEDS_UPDATE`, while the CLI transport may return all three possible states.
```
amccloud: This is a good place to call out only api supports the `parentId` option. Also need to update `branches.mergeState(BranchDescriptor, string)` above

src/endpoints/Branches.js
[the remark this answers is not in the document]
amccloud: A note in the docs would help 👍

abstract-sdk.d.ts
johncoder: Glad you caught that! Always easy to overlook.

tommoor:
Lets get 'er merged and a release published, this is blocking for quick sync very soon

PR #236, opened 2020-03-18
fix: Respect transportMode for branches.mergeState

anthony-j-castro:
When testing this in the front end, I noticed that an API call was always fired when accessing a branch's merge state, even when passing `transportMode: ["cli"]`. This PR fixes the issue by including `RequestOptions` in the endpoint declaration.

I also changed the `parent` parameter to `parentId`. It felt more intuitive to me, but if we don't want to make that change I can easily revert.


tommoor:
> I also changed the parent parameter to parentId. 

Seems like you fixed it to match the docs 😉

PR #239, opened 2020-03-24
feat: allow observing analytics on endpoints

rickharris:


rickharris:
@tommoor opened up a draft PR so I could get some feedback on the approach. It doesn't feel ideal but it's the simplest way I've found to achieve basically what I think we want. Here's what we have so far:

- You can pass an `analyticsCallback` option when initializing the client, and it will get called with a "type" string which is a combination of the endpoint and method (ex: `Activities#info`), the transport mode, and the time it took to complete.
- The timings happen inside of `Endpoint#apiRequest` and `Endpoint#cliRequest`
- Each endpoint will have to be updated to:
  - Set a `name` for each `Endpoint` implementation. We could possibly use `this.name` which is built into JS, but I'm pretty sure any minification would mangle those names.
  - Pass a `method` into every call of `apiRequest` or `cliRequest` so that we know if it's the `info`/`list`/etc functions being called on the endpoint.
    - I think this is the least palatable of the changes just because you have to pass it in everywhere and it basically does nothing except for analytics which might not even be used
    - The only less obtrusive way I can think to do this is to use `arguments.caller` inside of `Endpoint` but that method is deprecated and doesn't work in some places anymore.

tommoor:
> I think this is the least palatable of the changes 

Haha, I can't think of a better way of putting it. As you've described though I don't have any bright ideas to avoid this that don't get smashed by minification :(

src/endpoints/Activities.js
```diff
@@ -14,10 +14,13 @@ import Endpoint from "../endpoints/Endpoint";
 import { wrap } from "../util/helpers";
 
 export default class Activities extends Endpoint {
+  name = "Activities";
```
tommoor: I'd say we lower case these

src/endpoints/Endpoint.js
```diff
@@ -92,7 +94,16 @@ export default class Endpoint {
     /* istanbul ignore next */
     logAPIRequest.enabled && logAPIRequest(args);
 
+    const start = new Date();
```
tommoor: Probably better to use `performance.now` for these timing calls:
https://developer.mozilla.org/en-US/docs/Web/API/Performance/now

rickharris:
superseded by https://github.com/goabstract/abstract-sdk/pull/242

PR #241, opened 2020-03-24
feat: Assets cli support

berezovskycom:
Hi, I've been working on adding up cli to Assets endpoint lately, I currently have an issue with export (a.k.a. raw).
The server sends assets by parts (like 1/6 is exported), and our SDK just squashes these JSON objects into a string and then reproduces an error (Unexpected token { in JSON at position ...).
Do we have any already made solutions? Might `cursorPromise` help?
P.S. I don't want to merge right now, just wanted to ask a question

### DIFF
--- a/src/endpoints/Assets.js
+++ b/src/endpoints/Assets.js
@@ -134,7 +134,6 @@ export default class Assets extends Endpoint {
           `--urls=${latestDescriptor.url}`,
           `--filenames=${latestDescriptor.filename}`,
           `--output=${latestDescriptor.output || "asset"}`,
-          `--sha=${latestDescriptor.sha}`,
           latestDescriptor.expand ? `--expand` : ``
         ]);
 
--- a/src/endpoints/Assets.js
+++ b/src/endpoints/Assets.js
@@ -5,11 +5,13 @@ import { isNodeEnvironment, wrap } from "../util/helpers";
 import type {
   Asset,
   AssetDescriptor,
+  AssetHasChanges,
   BranchCommitDescriptor,
   FileDescriptor,
   ListOptions,
   RawOptions,
-  RequestOptions
+  RequestOptions,
+  AssetGenerateProgress
 } from "../types";
 import Endpoint from "../endpoints/Endpoint";
 
@@ -51,9 +53,6 @@ export default class Assets extends Endpoint {
 
   async file(descriptor: FileDescriptor, options: ListOptions = {}) {
     const { limit, offset, ...requestOptions } = options;
-    const latestDescriptor = await this.client.descriptors.getLatestDescriptor(
-      descriptor
-    );
 
     return this.createCursor<Promise<Asset[]>>(
       (nextOffset = offset) => ({
@@ -67,19 +66,6 @@ export default class Assets extends Endpoint {
           return this.apiRequest(
             `projects/${descriptor.projectId}/branches/${descriptor.branchId}/files/${descriptor.fileId}/assets?${query}`
           );
-        },
-        cli: async () => {
-          const response = await this.cliRequest([
-            "assets",
-            "export",
-            latestDescriptor.fileId,
-            `--branch-id=${latestDescriptor.branchId || "master"}`,
-            `--project-id=${latestDescriptor.projectId}`,
-            `--sha=${latestDescriptor.sha || "latest"}`
-          ]);
-
-          console.log(typeof response);
-          return wrap(response, response);
         },
         requestOptions
       }),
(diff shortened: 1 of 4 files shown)


src/endpoints/Assets.js
```diff
@@ -106,7 +126,67 @@ export default class Assets extends Endpoint {
...
+        ]);
+
+        return wrap(response, response);
+      },
+
       requestOptions
     });
   }
+
+  async fileChanged(descriptor: FileDescriptor, options: ListOptions = {}) {
```
tommoor: I think the pragmatic solution for now is to leave off this endpoint, we should only need the read endpoints to satisfy the needs of the internal ticket. Because this has taken a while to get to it's blocking the internal initiative now, so better to do less and get it landed. Make sense?

src/types.js
```diff
@@ -168,7 +168,13 @@ export type UserDescriptor = {|
 
 export type AssetDescriptor = {|
   assetId: string,
-  projectId: string
+  projectId: string,
```
tommoor: This type shouldn't need changing, it describes how to query a specific asset. If you want to _query_ assets for a specific branch / commit then you would use a BranchDescriptor / CommitDescriptor.

src/endpoints/Assets.js
```diff
@@ -14,14 +14,15 @@ import type {
 import Endpoint from "../endpoints/Endpoint";
 
 export default class Assets extends Endpoint {
-  info(descriptor: AssetDescriptor, requestOptions: RequestOptions = {}) {
+  async info(descriptor: AssetDescriptor, requestOptions: RequestOptions = {}) {
```
tommoor: No need for the `async` annotation here

src/endpoints/Assets.js
```diff
@@ -64,14 +68,30 @@ export default class Assets extends Endpoint {
...
+          const response = await this.cliRequest([
+            "assets",
+            "export",
+            latestDescriptor.fileId,
+            `--branch-id=${latestDescriptor.branchId || "master"}`,
+            `--project-id=${latestDescriptor.projectId}`,
+            `--sha=${latestDescriptor.sha || "latest"}`
+          ]);
+
+          console.log(typeof response);
```
tommoor: Logging.

src/endpoints/Assets.js
```diff
@@ -106,7 +126,67 @@ export default class Assets extends Endpoint {
...
+        ]);
+
+        return wrap(response, response);
+      },
+
+      requestOptions
+    });
+  }
+
+  async hasChanges(descriptor: AssetDescriptor, options: RawOptions = {}) {
```
tommoor:
```suggestion
  async hasChanges(descriptor: BranchCommitDescriptor, options: RawOptions = {}) {
```

src/endpoints/Assets.js
```diff
@@ -106,7 +126,67 @@ export default class Assets extends Endpoint {
 
         return arrayBuffer;
       },
+
+      cli: async () => {
+        const response = await this.cliRequest([
+          "assets",
+          "download",
+          `--urls=${latestDescriptor.url}`,
+          `--filenames=${latestDescriptor.filename}`,
```
tommoor: `filename`, `output`, `expand` etc should not be part of the descriptor – think of the descriptor more like an sql query. These should probably be in `RawOptions`, in fact `filename` already is.

What does `expand` do btw?

src/endpoints/Assets.js
```diff
@@ -48,8 +49,11 @@ export default class Assets extends Endpoint {
     });
   }
 
-  file(descriptor: FileDescriptor, options: ListOptions = {}) {
+  async file(descriptor: FileDescriptor, options: ListOptions = {}) {
```
tommoor: Have you confirmed that the output for cli/api transports is identical?

src/endpoints/Assets.js
```diff
@@ -106,7 +126,67 @@ export default class Assets extends Endpoint {
 
         return arrayBuffer;
       },
+
+      cli: async () => {
```
tommoor: Have you confirmed that the output for cli/api transports is identical?

src/endpoints/Assets.js
berezovskycom: Yes, totally.
Could you please send me a link an internal ticket somewhere?
I didn't know there was one

src/endpoints/Assets.js
berezovskycom: Thanks for noticing that, definitely makes sense.
According to cli command description `expand` downloads individual files instead of an archive.

src/endpoints/Assets.js
berezovskycom: Got it.

berezovskycom:
@tommoor Thanks for your feedback, without feedback people don't grow.
I fixed all the things you've mentioned, everything makes sense to me 🙌

src/endpoints/Assets.js
tommoor: https://goabstract.atlassian.net/browse/PLATFORM-891

PR #242, opened 2020-03-24
feat: allow observing analytics on endpoints

rickharris:
Adds a callback option to `Client` so you can receive data about each endpoint called in the sdk which will enable us to compare execution times for api vs. cli requests for quick sync:

image

I'll leave some comments inline.
Part of https://goabstract.atlassian.net/browse/PLATFORM-1029

### DIFF
--- a/abstract-sdk.d.ts
+++ b/abstract-sdk.d.ts
@@ -1826,9 +1826,17 @@ type AccessTokenOption =
   | (() => AccessToken) // TODO: Deprecate
   | (() => Promise<AccessToken>);
 
+type EndpointAnalytic = {
+  duration: number,
+  endpoint: string,
+  request: string,
+  transportMode: "api" | "cli"
+};
+
 type CommandOptions = {
   accessToken: AccessTokenOption,
   apiUrl: string | Promise<string>,
+  analyticsCallback: (analytics: EndpointAnalytic) => any,
   assetUrl: string | Promise<string>,
   previewUrl: string | Promise<string>,
   transportMode: ("api" | "cli")[],
--- a/abstract-sdk.d.ts
+++ b/abstract-sdk.d.ts
@@ -1826,7 +1826,7 @@ type AccessTokenOption =
   | (() => AccessToken) // TODO: Deprecate
   | (() => Promise<AccessToken>);
 
-type EndpointAnalytic = {
+type EndpointMetric = {
   duration: number,
   endpoint: string,
   request: string,
@@ -1836,7 +1836,7 @@ type EndpointAnalytic = {
 type CommandOptions = {
   accessToken: AccessTokenOption,
   apiUrl: string | Promise<string>,
-  analyticsCallback: (analytics: EndpointAnalytic) => any,
+  analyticsCallback: (analytics: EndpointMetric) => any,
   assetUrl: string | Promise<string>,
   previewUrl: string | Promise<string>,
   transportMode: ("api" | "cli")[],
--- a/src/endpoints/Activities.js
+++ b/src/endpoints/Activities.js
@@ -14,8 +14,9 @@ import Endpoint from "../endpoints/Endpoint";
 import { wrap } from "../util/helpers";
 
 export default class Activities extends Endpoint {
+  name = "activities";
   info(descriptor: ActivityDescriptor, requestOptions: RequestOptions = {}) {
-    return this.configureRequest<Promise<Activity>>({
+    return this.configureRequest<Promise<Activity>>("info", {
       api: async () => {
         const response = await this.apiRequest(
           `activities/${descriptor.activityId}`
@@ -33,6 +34,7 @@ export default class Activities extends Endpoint {
     const { limit, offset, ...requestOptions } = options;
 
     return this.createCursor<CursorPromise<Activity[]>>(
+      "list",
       (nextOffset = offset) => ({
         api: () => {
           const query = querystring.stringify({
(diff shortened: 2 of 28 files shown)


src/endpoints/Activities.js
```diff
@@ -33,6 +34,7 @@ export default class Activities extends Endpoint {
     const { limit, offset, ...requestOptions } = options;
 
     return this.createCursor<CursorPromise<Activity[]>>(
+      "list",
```
rickharris: There's a lot of noise in this PR because each endpoint was updated to including naming for endpoints and requests for the analytics. If you've seen one, you've seen them all. You could ignore the other endpoints in the diff (I'll point out anything that isn't this kind of change) unless you just wanted to double check that I didn't make a typo!

abstract-sdk.d.ts
```diff
@@ -1826,9 +1826,17 @@ type AccessTokenOption =
...
+  duration: number,
+  endpoint: string,
+  request: string,
+  transportMode: "api" | "cli"
+};
+
 type CommandOptions = {
   accessToken: AccessTokenOption,
   apiUrl: string | Promise<string>,
+  analyticsCallback: (analytics: EndpointAnalytic) => any,
```
rickharris: 🍖

tests/endpoints/Endpoint.test.js
```diff
@@ -0,0 +1,80 @@
...
+
+      await endpoint.info();
+
+      if (!analyticsResult) {
+        throw new Error("analytics should be defined");
+      }
+      expect(analyticsResult.duration).toBeGreaterThan(0);
+      expect(analyticsResult.endpoint).toEqual("exampleEndpoint");
+      expect(analyticsResult.request).toEqual("info");
+      expect(analyticsResult.transportMode).toEqual("cli");
```
rickharris: 🍖 see this test file for some example usage.

src/endpoints/Activities.js
amccloud: This is not the first time we've needed this. At one point in time the entire interface of `Client` was a `Proxy` object so that we could see this information at runtime. Not sure what happened to the Proxy, but kinda glad to see it is gone as it was only used for error messages

Can't really think of a nice way to do this without using [babel macros](https://github.com/kentcdodds/babel-plugin-macros). Could create a macro that extracts `constructor.name/arguemnts.callee`  into a string before it gets mangled

rickharris: `constructor.name` could get us part of the way there, but `arguments.callee` is deprecated and already does not work in some environments (including our test environment)

abstract-sdk.d.ts
```diff
@@ -1826,9 +1826,17 @@ type AccessTokenOption =
   | (() => AccessToken) // TODO: Deprecate
   | (() => Promise<AccessToken>);
 
+type EndpointAnalytic = {
```
tommoor: Analytic is not the singular of "Analytics", perhaps "Metric"? We've used that before in the `ui` codebase to describe a value returned from the profiler.

src/endpoints/Endpoint.js
```diff
@@ -51,11 +52,22 @@ export default class Endpoint {
...
-          if (!config[mode]) {
+          const request = config[mode];
+          if (!request) {
             throw new EndpointUndefinedError(mode);
           }
-          const operation = config[mode].call(this);
+          const start = performance.now();
+          const operation = request.call(this);
           response = await operation;
+          const end = performance.now();
```
tommoor: Just a note that if the request fails then we won't report metrics for it, I think that's fine but just wanted to call that out incase it was accidental.

abstract-sdk.d.ts
rickharris: renamed to "EndpointMetric"

src/endpoints/Endpoint.js
rickharris: This was intentional. Logging times for errors didn't seem important for what we're looking for here.

tests/endpoints/Endpoint.test.js
```diff
@@ -0,0 +1,80 @@
...
+        return sleep(1);
+      }
+    });
+  }
+}
+
+describe("endpoint", () => {
+  describe("analyticsCallback", () => {
+    test("api", async () => {
+      let analyticsResult: ?EndpointMetric;
```
amccloud: nit: If you pass `done`  to the `test` callback you can move all of your assertions into the analytics callback and call `done()` at the end

tests/endpoints/Endpoint.test.js
```diff
@@ -0,0 +1,80 @@
...
+      const options = {
+        ...CLIENT_CONFIG,
+        analyticsCallback: analytics => (analyticsResult = analytics)
+      };
+
+      const client = new Client(options);
+      const endpoint = new ExampleEndpoint(client, {
+        ...options,
+        transportMode: ["cli"],
+        webUrl: "https://example.com"
```
amccloud: webUrl is required in both places?

rickharris: yeah it's a flow error if you don't pass it in here.

amccloud: Oh right i'm thinking this is a Client. Ignore me 🙃

tests/endpoints/Endpoint.test.js
rickharris: what happens if `done` is never called (like, if we screwed something up and the callback is never called)?

rickharris: it seems to wait 5 seconds before it fails the test. I'm ok with doing it this way if it means not slowing down the test that much.

amccloud: It’s there for this case. I don’t mind either way though

PR #245, opened 2020-04-05
feat: Add func generators support

berezovskycom:
Function Generators are not available in IE11, and since we want to support it, we shall include function generators when transpiling.


tommoor:
The meat of this makes sense to me, could you clean up the PR so it's mergable?

PR #248, opened 2020-04-09
feat: allow analytics callback to be set on Client

rickharris:
This is a follow-up to https://github.com/goabstract/abstract-sdk/pull/242. We have a need to be able to set the analytics callback separately between desktop and web, so we need to be able to set it outside of the constructor. This PR maintains the same ability to pass it as part of the constructor, but also be able to set the callback directly on `Client`.

### DIFF
--- a/src/Client.js
+++ b/src/Client.js
@@ -51,7 +51,7 @@ export default class Client {
   users: Users;
   webhooks: Webhooks;
 
-  analyticsCallback: ?AnalyticsCallback;
+  _analyticsCallback: ?AnalyticsCallback;
 
   constructor(options: $Shape<CommandOptions> = {}) {
     options = {
@@ -65,7 +65,7 @@ export default class Client {
       ...options
     };
 
-    this.analyticsCallback = options.analyticsCallback;
+    this._analyticsCallback = options.analyticsCallback;
     this.activities = new Activities(this, options);
     this.assets = new Assets(this, options);
     this.branches = new Branches(this, options);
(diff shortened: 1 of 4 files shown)


amccloud:
> We have a need to be able to set the analytics callback separately between desktop and web, so we need to be able to set it outside of the constructor

Can you say more?

```js
const abstract = new Abstract.Client({
  analyticsCallback: () => {
    if (isWeb) {
      // web
    } else {
      // not web
    }
  }
});
```

Would that work?

rickharris:
Yeah, sorry I didn't go into as much detail because it involves specifics about the Abstract app. If we initialize the callback in `core`, then we have to use dependency injection around the redux store, and the injection is causing bad require loops. You can see how we're using this in https://github.com/goabstract/ui/pull/2847

src/Client.js
```diff
@@ -51,6 +51,8 @@ export default class Client {
   users: Users;
   webhooks: Webhooks;
 
+  analyticsCallback: ?AnalyticsCallback;
```
amccloud, line 54: > then we have to use dependency injection around the redux store, and the injection is causing bad require loops

@rickharris gotcha, yea that is annoying. In that case, can we keep this property private by prefixing it with `_`?

```suggestion
  _analyticsCallback: ?AnalyticsCallback;
```

rickharris: Yeah, sure, if that's the pattern we want to keep. I wasn't really planning on documenting this or anything 😆. The `_` doesn't do anything special right? People will still see it if they inspect the client object?

amccloud: Yup, just for the sdk as one of it's features is a public interface. `_` helps signal this is not documented and this will likely change without notice

PR #247, opened 2020-04-07
fix: Added missing type BranchMergeState

mascardoso:

PR #251, opened 2020-04-14
fix: Add perfomance from  perf_hooks (Node v8.5.0+)

berezovskycom:
Fixes the error from `8.0.0-beta.20` regarding:

```
api: ReferenceError: performance is not defined
at makeRequest (project/node_modules/abstract-sdk/dist/endpoints/Endpoint.js:79:25)
```

### DIFF
--- a/src/endpoints/Endpoint.js
+++ b/src/endpoints/Endpoint.js
@@ -3,6 +3,8 @@
 import { Readable } from "stream";
 import "cross-fetch/polyfill";
 import { spawn } from "child_process";
+// $FlowFixMe
+import { performance } from "perf_hooks";
 import uuid from "uuid/v4";
 import { version } from "../../package.json";
 import Client from "../Client";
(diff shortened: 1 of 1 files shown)


berezovskycom:
Let's push [this 247 first](https://github.com/goabstract/abstract-sdk/pull/247) and then merge this one

src/endpoints/Endpoint.js
```diff
@@ -3,6 +3,8 @@
 import { Readable } from "stream";
 import "cross-fetch/polyfill";
 import { spawn } from "child_process";
+// $FlowFixMe
+import { performance } from "perf_hooks";
```
tommoor, line 7: I haven't ran this, but how does it handle running in the browser? It seems like you're swapping problems in one environment for another. I'd suggest creating a small wrapper and only `requiring` "perf_hooks" if you're not in a browser.

rickharris:
Looking back on this I think the mistake I made was putting the perf timings in the code unconditionally. I wish that I had put them inside the analytics callback check and checked for the existence of `window.performance.now` as well before calling it.

Since this isn't a documented feature, the only reason people are seeing this is because the calls to `performance.now` happen for _everyone_ rather than just for folks who have configured the `analyticsCallback`.

PR #252, opened 2020-04-14
fix: only use performance.now when it is available

rickharris:
Proposing an alternate solution to https://github.com/goabstract/abstract-sdk/pull/251, since we're not sure if node's `perf_tools` module is going to work in the browser. We know that this will give us what we need for the internal analytics reporting, and doing it this way won't break for node users. If we need to make this public in the future, we can find a way to use either `perf_tools` or `window.performance` depending on whether the consumer is a browser or a node environment, but I think this is probably what we need for now.


berezovskycom:
Makes sense. Thanks @rickharris!

PR #256, opened 2020-04-16
chore: use api version 25 for collection lists

rickharris:
The API including `branches` as part of the `collections#index` response is being put behind an API version (https://github.com/goabstract/projects/pull/4521). This PR bumps the API version for that call to the API.

### DIFF
--- a/src/endpoints/Collections.js
+++ b/src/endpoints/Collections.js
@@ -112,7 +112,7 @@ export default class Collections extends Endpoint {
           `projects/${projectId}/collections?${query}`,
           {
             headers: {
-              "Abstract-Api-Version": requestOptions._version || API_VERSION
+              "Abstract-Api-Version": requestOptions._version || 25 // 25 includes `branches` as part of the response
             }
           }
         );


src/endpoints/Collections.js
```diff
@@ -112,7 +112,7 @@ export default class Collections extends Endpoint {
           `projects/${projectId}/collections?${query}`,
           {
             headers: {
-              "Abstract-Api-Version": requestOptions._version || API_VERSION
```
rickharris, line 115: I have no idea why we're attempting to use the same version for multiple API endpoints. Seems like a source of bugs.

src/endpoints/Collections.js
```diff
@@ -112,7 +112,7 @@ export default class Collections extends Endpoint {
...
+              "Abstract-Api-Version": requestOptions._version || 25 // 25 includes `branches` as part of the response
```
timc1, line 115: We need to update the version here: https://github.com/goabstract/ui/blob/4343eaddab383873ad6fb17e8c5ec01bcb4c7033/core/src/requests/collections.js#L146

rickharris: good call

timc1: + 1 more since it's the project collections we need to update it in `ProjectCollectionsFetchRequest`

rickharris: @timc1 not sure I understand that last comment.

Also looking at the original link, I'm noticing that is for `CollectionFetchRequest` (singular), so I don't think it should be affected by this.

timc1: Yeah my brain is fried. the reason i thought about that was because the filtering PR was specifically for project collections. Ignore this!

timc1:
@rickharris do you think we should play it safe with defaulting to version 16? https://github.com/goabstract/abstract-sdk/blob/dacff70a2b739fd6834821a1f6487a51cbe99240/src/endpoints/Collections.js#L17

The only place we currently need version 25 is in the `ProjectCollections` page, which needs the `branches` in the collections response.

rickharris:
No, I don't think so. I'm not sure what would be more safe about that. The whole point of the API version is so that we can pair the API response with code that is capable of handling it. So once @tommoor's PR that uses the correct branch head is in, it won't matter which collection endpoints return `branches` in the response.

timc1:
> No, I don't think so. I'm not sure what would be more safe about that. The whole point of the API version is so that we can pair the API response with code that is capable of handling it. So once @tommoor's PR that uses the correct branch head is in, it won't matter which collection endpoints return `branches` in the response.

Sounds good - looks like 16 itself is fine: https://github.com/goabstract/projects/blob/8709bbf55964cd269f71d4a4d89f3a71484792cf/stable/api/app/controllers/collections_controller.rb#L66

PR #259, opened 2020-04-27
fix: Optional Error.captureStackTrace

amccloud:
- Do not try to capture stack trace when `Error.captureStackTrace` is not available

Fixes https://sentry.io/organizations/elastic-projects/issues/1633606217/?project=103885

### DIFF
--- a/src/errors.js
+++ b/src/errors.js
@@ -10,6 +10,7 @@ export class BaseError extends Error {
     super(message);
     this.name = this.constructor.name;
 
+    /* istanbul ignore next */
     if (Error.captureStackTrace) {
       Error.captureStackTrace(this, this.constructor); // This is a Node API that helps stack trace readability
     }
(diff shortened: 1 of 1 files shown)


src/errors.js
```diff
@@ -9,7 +9,10 @@ export class BaseError extends Error {
   constructor(message: string) {
     super(message);
     this.name = this.constructor.name;
-    Error.captureStackTrace(this, this.constructor);
+
+    if (Error.captureStackTrace) {
```
tommoor, line 13: This is a nodejs only method? Lets add a quick note to this extent

rickharris: This comment from the PR would be a code comment probably https://github.com/goabstract/abstract-sdk/pull/82#discussion_r250804405

amccloud:
Really not sure what is going on with the test for this PR. The test pass but exit with 1 🤔

PR #266, opened 2020-05-04
refactor: moves url param to ShareUrlDescriptor

enriquenov:
In order to access the `shareId` prop in the `ui` project, we need to remove its union to `{| url: string |}` in the `ShareDescriptor` type. The url part was moved to a brand new `ShareUrlDescriptor`.

Will be updating accordingly the `ui` project right after this gets approved starting with this [PR](https://github.com/goabstract/ui/pull/2985).

### DIFF
--- a/docs/abstract-api.md
+++ b/docs/abstract-api.md
@@ -2267,7 +2267,13 @@ Reference for the parameters required to load resources with the Abstract SDK.
 ### ShareDescriptor
 
 ```js
-{ url: string } | { shareId: string }
+{ shareId: string }
+```
+
+### ShareUrlDescriptor
+
+```js
+{ url: string }
 ```
 
 ### BranchDescriptor
--- a/docs/abstract-api.md
+++ b/docs/abstract-api.md
@@ -1834,7 +1834,7 @@ A share is a shareable url to an object in Abstract. You can use the desktop or 
 
 ![API][api-icon]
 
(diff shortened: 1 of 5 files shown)


enriquenov:
@amccloud you got it! Thank you 👍

docs/abstract-api.md
```diff
@@ -2267,7 +2267,13 @@ Reference for the parameters required to load resources with the Abstract SDK.
 ### ShareDescriptor
 
 ```js
-{ url: string } | { shareId: string }
+{ shareId: string }
+```
+
+### ShareUrlDescriptor
```
anthony-j-castro, line 2273: Just jumping in because I noticed this while working on my PR: I think you'll have to add `ShareUrlDescriptor` to the "Load info for a share" and "Using share.descriptor" sections?

enriquenov: good catch, thanks!

docs/abstract-api.md
```diff
@@ -1854,12 +1854,12 @@ abstract.shares.info({
...
 ![API][api-icon]
 
-`shares.info(ShareDescriptor, RequestOptions): Promise<Share>`
+`shares.info(ShareDescriptor | ShareUrlDescriptor, RequestOptions): Promise<Share>`
 
 List all files for branch's share url
 
 ```js
-const branchShare = await abstract.share.info({
+const branchShare = await abstract.shares.info({
```
enriquenov, line 1862: typo

PR #267, opened 2020-05-04
feat: Add ability to set a shareId

anthony-j-castro:
We recently ran into an issue where some requests on private share link pages would fail because we were only allowing a single `accessToken`. For these share links, `accessToken` was being set to the share ID, and the authentication header would never get configured.

This PR adds a separate `shareId` option that can be set in addition to the `accessToken`, allowing both headers to exist on a request.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -41,6 +41,10 @@ _Default value: `https://previews.goabstract.com`_
 
 This option can be used to specify a custom URL that points to an instance of the Abstract preview service. This is used when building preview image URLs used by the [API transport](/docs/transports). This option is useful if HTTP requests should be routed locally or through a proxy server.
 
+### `shareId`
+
+This option can be used to pass a share identifier (`string`, [`ShareDescriptor`](/docs/abstract-api/#sharedescriptor), or [`ShareUrlDescriptor`](/docs/abstract-api/#shareurldescriptor)) to access objects associated with a public [share](/docs/abstract-api/#shares) without having to authenticate with an `accessToken`.
+
 ### `transportMode`
 
 _Default value: `["api"]`_
--- a/src/endpoints/Endpoint.js
+++ b/src/endpoints/Endpoint.js
@@ -265,6 +265,17 @@ export default class Endpoint {
         ? { Authorization: `Bearer ${token}` }
         : { "Abstract-Share-Id": token && inferShareId(token) };
 
+    let shareHeader;
+    const shareToken = this.options.shareToken
+      ? await this.options.shareToken()
+      : undefined;
+    if (shareToken) {
+      shareHeader = {
+        "Abstract-Share-Id":
+          typeof shareToken === "string" ? shareToken : inferShareId(shareToken)
+      };
+    }
+
     const headers = {
       Accept: "application/json",
       "Content-Type": "application/json",
@@ -272,6 +283,7 @@ export default class Endpoint {
       "X-Amzn-Trace-Id": `Root=1-${new Date().getTime()}-${uuid()}`,
       "Abstract-Api-Version": "8",
       ...tokenHeader,
+      ...shareHeader,
       ...customHeaders
     };
 
--- a/src/endpoints/Endpoint.js
+++ b/src/endpoints/Endpoint.js
@@ -259,11 +259,11 @@ export default class Endpoint {
   }
 
   async _getFetchHeaders(customHeaders?: { [key: string]: string }) {
-    const token = await this._getAccessToken();
-    const tokenHeader =
-      typeof token === "string"
-        ? { Authorization: `Bearer ${token}` }
-        : { "Abstract-Share-Id": token && inferShareId(token) };
+    let accessTokenHeader;
+    const accessToken = await this._getAccessToken();
+    if (accessToken) {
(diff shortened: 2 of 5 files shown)


anthony-j-castro:
@amccloud this is still a WIP but I wanted to get your opinion on this approach before I finish things up with tests and documentation (and having to change some things after https://github.com/goabstract/abstract-sdk/pull/266 is merged).

src/endpoints/Endpoint.js
```diff
@@ -265,13 +265,25 @@ export default class Endpoint {
         ? { Authorization: `Bearer ${token}` }
         : { "Abstract-Share-Id": token && inferShareId(token) };
```
amccloud, line 266: ☝️ notice above we have `Abstract-Share-Id` already

anthony-j-castro:
@amccloud is this on the right track? I removed the ability for `accessToken` to accept a share value--now it has to be explicitly sent in as `shareId`.

src/types.js
```diff
@@ -153,7 +161,7 @@ export type EndpointMetric = {
 export type AnalyticsCallback = EndpointMetric => mixed;
 
 export type CommandOptions = {
-  accessToken: AccessTokenOption,
+  ...AuthenticationOptions,
```
amccloud, line 164: If this type is not used anywhere else we can just inline the properties here

src/endpoints/Endpoint.js
```diff
@@ -259,19 +259,29 @@ export default class Endpoint {
   }
 
   async _getFetchHeaders(customHeaders?: { [key: string]: string }) {
-    const token = await this._getAccessToken();
-    const tokenHeader =
-      typeof token === "string"
-        ? { Authorization: `Bearer ${token}` }
-        : { "Abstract-Share-Id": token && inferShareId(token) };
```
anthony-j-castro, line 266: We're removing the ability to pass a share ID as an `accessToken`. Technically this would be a breaking change, but it's not documented for the public and I'm going to update our only usage in ui after a prerelease is cut.

anthony-j-castro:
@amccloud I've been testing this locally using a linked version and it's working well 👍

src/types.js
```diff
@@ -137,12 +137,15 @@ export type CollectionsListOptions = {
...
 };
 
-export type AccessToken = ?string | ShareDescriptor;
+export type AccessToken = ?string;
 export type AccessTokenOption =
   | AccessToken // TODO: Deprecate?
   | (() => AccessToken) // TODO: Deprecate
   | (() => Promise<AccessToken>);
 
+export type ShareIdentifier = ?string | ShareDescriptor;
```
amccloud, line 146: I think the maybe type here will allow `shareId: null`

```suggestion
export type ShareIdentifier = string | ShareDescriptor;
```

anthony-j-castro: We actually `null` out `shareToken` in ui. So when we call `getShareToken`, it could technically be `null`. If we make this change then I guess I would just pass in an empty string if `shareToken` is `null`?

amccloud: @anthony-j-castro that makes sense. In that case we could omit the `shareId` option when `getShareToken` is null

Can cast null to undefined with `getShareToken() || undefined` or can update getShareToken to return `undefined` initially. Since getShareToken is synchronous, If it's `undefined` it won't ever be defined

> I would just pass in an empty string if shareToken is null?

Personally I avoid `""` outside of a default value for user input. It seems to introduce invalid states that make it harder to reason about control flow - as you lose the ability to distinguish between the states: `string | null | undefined`

 With `""`, null and undefined are collapsed into `string` via `""` and that is likely not the expected string so `string` loses meaning too

src/types.js
```diff
@@ -137,12 +137,15 @@ export type CollectionsListOptions = {
...
 };
 
-export type AccessToken = ?string | ShareDescriptor;
+export type AccessToken = ?string;
 export type AccessTokenOption =
   | AccessToken // TODO: Deprecate?
   | (() => AccessToken) // TODO: Deprecate
   | (() => Promise<AccessToken>);
 
+export type ShareIdentifier = string | ShareDescriptor;
```
amccloud, line 146: nit: 

```suggestion
type ShareIdentifier = string | ShareDescriptor;
```

or just inline `string | ShareDescriptor` below. We should be selective about which types we introduce to the public api

anthony-j-castro: Went ahead and inlined it. I'll also make sure to update this after Enrique's PR lands.

docs/configuration.md
```diff
@@ -41,6 +41,10 @@ _Default value: `https://previews.goabstract.com`_
 
 This option can be used to specify a custom URL that points to an instance of the Abstract preview service. This is used when building preview image URLs used by the [API transport](/docs/transports). This option is useful if HTTP requests should be routed locally or through a proxy server.
 
+### `shareId`
+
+This option can be used to pass a share identifier (`string`, [`ShareDescriptor`](/docs/abstract-api/#sharedescriptor), or [`ShareUrlDescriptor`](/docs/abstract-api/#shareurldescriptor)) to access objects associated with a public [share](/docs/abstract-api/#shares) without having to authenticate with an `accessToken`.
+
```
anthony-j-castro, line 47: @amccloud here's my first attempt at documenting. Suggestions are welcome 🙃

PR #269, opened 2020-05-07
fix: Correctly handle error when CLI response is unparsable

tommoor:
It is unlikely, but possible for the CLI to respond with something that is not JSON-parsable. Currently if this happens an uncaught error is thrown. In our applications this would result in a request that is never resolved.

This change makes sure that the error is caught and the promise is properly rejected.

image


tommoor:
I'm not sure what's going on with CI in this project – the build passed...

PR #270, opened 2020-05-07
feat: Allow share and auth tokens

berezovskycom:
Had to make a pr again because of an error in flow.
Will fix that.


berezovskycom:
https://github.com/goabstract/abstract-sdk/pull/271

PR #275, opened 2020-05-13
Feat: list files from a commit

berezovskycom:


tests/endpoints/Files.test.js
```diff
@@ -98,6 +98,32 @@ describe("files", () => {
       ]);
     });
 
+    test("api - commit", async () => {
+      mockAPI(
+        "/projects/project-id/branches/branch-id/commits/commit-id/files",
```
berezovskycom, line 103: Heyo @amccloud! Could you please describe how current testing works? E.g. what project-id, or branch-id it uses?

abstract-sdk.d.ts
```diff
@@ -393,7 +393,8 @@ type ObjectDescriptor = {
 type BranchCommitDescriptor = {
   sha: string,
   projectId: string,
-  branchId: string | "master"
+  branchId: string | "master",
+  commitId?: string | "latest"
```
berezovskycom, line 397: Hi @konstantinn! One question: do we handle "latest" value of commit id on the back end side?

tests/endpoints/Files.test.js
berezovskycom: @Nivl hi! maybe you can help me? What project-id and branch-id gets used in this case?

amccloud: @berezovskycom Keep in mind this is no a end-to-end test of backend services and such. The responsibility of the sdk is to make sure it calls the api correctly and formats the returned data correctly. We accomplish this test by mocking the expected request from the sdk with the expected response

As a convention we use the name of the property is used as a placeholder value for testing. `branch-id` and `project-id` are valid ids and are a bit easier to debug and understand during test failures compared to random UUIDs

src/endpoints/Files.js
```diff
@@ -59,10 +59,16 @@ export default class Files extends Endpoint {
...
           `projects/${latestDescriptor.projectId}/branches/${latestDescriptor.branchId}/files`
         );
 
+        if (latestDescriptor.commitId) {
+          response = await this.apiRequest(
+            `projects/${latestDescriptor.projectId}/branches/${latestDescriptor.branchId}/commits/${latestDescriptor.commitId}/files`
+          );
+        }
+
         return wrap(response.files, response);
```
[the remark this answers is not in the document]
berezovskycom: Yup, makes sense

abstract-sdk.d.ts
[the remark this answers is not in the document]
berezovskycom: I think it's different when it gets used in Files.list(), since `sha` is used for files in cli and not commit. We can create another descriptor for clearer purpose.

abstract-sdk.d.ts
```diff
@@ -394,7 +394,7 @@ type BranchCommitDescriptor = {
   sha: string,
   projectId: string,
   branchId: string | "master",
-  commitId?: string | "latest"
```
berezovskycom: `latest` will not work, because back end side can't decide what is the latest commit

amccloud: @berezovskycom this is correct. We never pass `"latest"` to our backend as it is not cache-friendly. We do have the ability to resolve "latest" to a real sha or commit id client-side using [getLatestDescriptor](https://github.com/goabstract/abstract-sdk/search?q=getLatestDescriptor&unscoped_q=getLatestDescriptor)

abstract-sdk.d.ts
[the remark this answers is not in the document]
amccloud: @berezovskycom We've used `sha`, `commitId`, `xxxSha` interchangeably.  If you're referencing something outside of a commit on a branch, an additional descriptor would be preferred over optional 👍 

It helps calls out situations like what happens if `sha` and `commitId` are defined together? Should they both be defined at the same time and if so, are they always the same value when they are both defined?

In this case I believe we only need one or the other, so `sha` and `commitId` may be aliases?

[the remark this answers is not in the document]
berezovskycom: @amccloud I see where you going, makes sense.

berezovskycom:
@amccloud any idea why tests fail? Basically what we do is just change the req url. I did change url in tests, but now all of them fail. Have you been in such trouble before?

amccloud:
@berezovskycom no I haven't seen this error before. Looks like your additions is throwing an additional exception during some test causing the two exceptions to be rolled up into `MultiError`

```
 Expected constructor: FileExportError
 Received constructor: MultiError
 ```

If I had to guess it's because some testing mocks need to be updated that previously relied on the path you changed [here](https://github.com/goabstract/abstract-sdk/pull/275/files#diff-9736403df2ab7184dcf1488e084330d2L63)

berezovskycom:
@amccloud got it! Ready for merging.

PR #281, opened 2020-06-09
feat: add marketo analytics to documentation website

berezovskycom:
Adds Marketo analytics to webpages, so our marketing team can learn more about our customers.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
   "scripts": {
     "check-ts-types": "tsc abstract-sdk.d.ts --lib es2017,dom,esnext.asynciterable",
     "prepare": "./scripts/install-types.sh",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -32,7 +32,7 @@
   "scripts": {
     "check-ts-types": "tsc abstract-sdk.d.ts --lib es2017,dom,esnext.asynciterable",
     "prepare": "./scripts/install-types.sh",
-    "lint": "eslint . --ignore-path .gitignore",
+    "lint": "eslint . --ignore-path .gitignore --ignore-pattern /website/static/js/marketo.js",
```
amccloud, line 35: Their js is just javascript. Is it possible to just run it through prettier or `lint --fix` to avoid having to add this special case?

Issue #276, opened 2020-05-27
abstract-cli could not be accessed at the expected path


xdrdak:
When attempting to use `abstract-sdk@8.0.0-beta.30`, an error is thrown on client instantiation:

```
abstract-cli could not be accessed at the expected path [...]
```

This bug is reproducible under the following environment:

**OS**: macOS Catalina Version 10.15.4
**Node**: v14.0.0

---

Additional Details:

Bug cannot be reproduced on `node v12.17.0 lts` or `node v10.16.0`

slmosl:
Going over the `resolveCLIBinary` code in abstract-cli/index.js , it looks like there is a type mismatch when using `accessSync`

The way it is used right now in the code is `fs.accessSync(binPath, "rx");`
while the correct way to use it is `fs.accessSync(binPath, fs.constants.R_OK | fs.constants.X_OK)`

see: https://nodejs.org/api/fs.html#fs_fs_accesssync_path_mode

berezovskycom:
Hi @xdrdak @slmosl we'll make a release of abstract-cli asap.

PR #284, opened 2020-06-23
feat: update cli

berezovskycom:
Fixes an Error in node 14, regarding fs module.
Tested on node 10, 12, 14 and it works fine.
Issue #276

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
   },
   "dependencies": {
     "@babel/runtime-corejs3": "^7.9.6",
-    "@elasticprojects/abstract-cli": "^11.1.1",
+    "@elasticprojects/abstract-cli": "^11.1.0",
     "core-js": "^3.6.5",
     "cross-fetch": "^3.0.1",
     "debug": "^4.0.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -68,7 +68,7 @@
   },
   "dependencies": {
     "@babel/runtime-corejs3": "^7.9.6",
-    "@elasticprojects/abstract-cli": "^4.3.0",
+    "@elasticprojects/abstract-cli": "^11.1.1-dev.20200622100140",
```
tommoor, line 71: Can we use a non-dev release, `11.1.1` should contain the fix – do you know which PR on the backend addressed the issue?

berezovskycom: Oh, yeah, sure, my bad.
[#4854](https://github.com/goabstract/projects/pull/4854)
Does 11.1.1 contain the fix?

tommoor: It does

berezovskycom:
@tommoor had to downgrade to `11.1.0`, since `11.1.1 `is unavailable via npm (works anyway).
Screenshot 2020-06-24 at 15 09 07

PR #286, opened 2020-07-02
chore: Remove features from organizations response

tommoor:
In order to make API requests more performant we no longer return feature flags inside of organization payloads.

### DIFF
--- a/src/endpoints/Organizations.js
+++ b/src/endpoints/Organizations.js
@@ -7,6 +7,11 @@ import type {
 import Endpoint from "../endpoints/Endpoint";
 import { wrap } from "../util/helpers";
 
+// Version 27 does not return features for organizations
+const headers = {
+  "Abstract-Api-Version": "27"
+};
+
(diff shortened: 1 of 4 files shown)


berezovskycom:
Thanks for mentioning the #3305 pr, seems like a reasonable update to sdk. lgtm 👍

src/endpoints/Organizations.js
```diff
@@ -29,7 +37,9 @@ export default class Organizations extends Endpoint {
   list(requestOptions: RequestOptions = {}) {
     return this.configureRequest<Promise<Organization[]>>("list", {
       api: async () => {
-        const response = await this.apiRequest("organizations");
+        const response = await this.apiRequest("organizations", {
+          headers
+        });
```
amccloud, line 42: Did we also use to return features for list request?

tommoor: Yep, anywhere an org was serialized

PR #287, opened 2020-07-06
chore: Add stream package as a dependency

dmuneras:
## Description

- Add stream as a dependency of the package. 

When using this package in a Ember Application I get the following error: 

Screenshot 2020-07-06 at 15 27 37

If I look at the file in the dust folder I can see the require to stream:

Screenshot 2020-07-06 at 15 37 49

## Note

It could be related to a specific configuration of the build process of the app, but, if it doesn't cause any harm for you, it will be cool to avoid the problem by adding the explicit dependency in the package.json.

Best regards,
Danie.

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,6 +74,7 @@
     "debug": "^4.0.1",
     "js-sha256": "^0.9.0",
     "query-string": "^6.1.0",
+    "stream": "0.0.2",
     "uuid": "^3.3.2"
   },
   "eslintConfig": {


tommoor:
I don't think adding the dependency for browser is the way forward here, `stream` is for the node module environment – we should probably move the import to being optional.

cc @amccloud

amccloud:
👍 

stream is only used for cli responses so it can be added to [this list](https://github.com/goabstract/abstract-sdk/blob/master/package.json#L26)

tommoor:
@dmuneras want to try that approach and see if it works in your particular environment?

dmuneras:
Thanks @tommoor and @amccloud for the super quick response. I will give it a try and let you know.

tommoor:
closed in favor of https://github.com/goabstract/abstract-sdk/pull/319

PR #289, opened 2020-07-09
feat(oauth): add OAuth endpoint with generateUrl() method

berezovskycom:
This PR is the 2nd part of adding OAuth2.0 into Abstract SDK.
[Link to OAUTH2.0 RFC](https://goabstract.atlassian.net/wiki/spaces/ENG/pages/341705615/OAuth+2.0)
[Link to OAuth interface for SDK](https://goabstract.atlassian.net/wiki/spaces/ENG/pages/447643698/RFC+OAuth+interface+for+JavaScript+SDK#Architecture-Summary)

With this feature, developers are able to generate url for authenticating users to their application.

Example:

```
const abstract = new Abstract.Client({
  clientId,
  clientSecret,
  redirectUri
});

const url = abstract.oauth.generateUrl({
  state: "somestate"
});
```

PR #290, opened 2020-07-10
feat: OAuth 2.0 functionality

berezovskyicom:
This PR adds OAuth2.0 into Abstract SDK.
[Link to OAUTH2.0 RFC](https://goabstract.atlassian.net/wiki/spaces/ENG/pages/341705615/OAuth+2.0)
[Link to OAuth interface for SDK](https://goabstract.atlassian.net/wiki/spaces/ENG/pages/447643698/RFC+OAuth+interface+for+JavaScript+SDK#Architecture-Summary)

Adds add `getToken()` for making a `POST` request to `auth/tokens` with `authorizationCode`.
And `setToken()` for storing returned `accessToken` in client.
`generateAuthorizeurl()` is used for returning an oauth url to Abstract.

Example:

```
let abstract = new Abstract.Client({
  clientId,
  clientSecret,
  redirectUri,
});

async function oAuth(code) {
  return await abstract.oauth
    .getToken({
      authorizationCode: code
    })
    .then(token => {
      abstract = abstract.oauth.setToken(token);

      return abstract;
    })
    .catch(error => console.log(error));
}

const authorizedAbstract = oAuth("authentication_code");

authorizedAbstract.then(() => {
  // use sdk as usual
});
```

Things to consider:
`client.oauth.setToken()` returns `new Client({})` with the same `options + accessToken`.
Is it really possible to change options for Client and all other endpoints without returning a new class?
I've tried a few ways, but the only `options` get changed are in `oauth` endpoint. Any thoughts?

### DIFF
--- a/abstract-sdk.d.ts
+++ b/abstract-sdk.d.ts
@@ -1858,7 +1858,7 @@ type CommandOptions = {
   clientSecret?: string,
   previewUrl: string | Promise<string>,
   shareId?: () => Promise<string | ShareDescriptor | ShareUrlDescriptor | void>,
-  redirectUri?: string | Promise<string>,
+  redirectUri?: string,
   transportMode: ("api" | "cli")[],
   webUrl: string | Promise<string>
 };
--- a/abstract-sdk.d.ts
+++ b/abstract-sdk.d.ts
@@ -1855,8 +1855,10 @@ type CommandOptions = {
   apiUrl: string | Promise<string>,
   analyticsCallback: AnalyticsCallback,
   assetUrl: string | Promise<string>,
+  clientSecret?: string | Promise<string>,
   previewUrl: string | Promise<string>,
   shareId?: () => Promise<string | ShareDescriptor | ShareUrlDescriptor | void>,
+  redirectUri?: string | Promise<string>,
   transportMode: ("api" | "cli")[],
   webUrl: string | Promise<string>
 };
--- a/abstract-sdk.d.ts
+++ b/abstract-sdk.d.ts
@@ -1855,7 +1855,7 @@ type CommandOptions = {
   apiUrl: string | Promise<string>,
   analyticsCallback: AnalyticsCallback,
   assetUrl: string | Promise<string>,
-  clientSecret?: string | Promise<string>,
+  clientSecret?: string,
   previewUrl: string | Promise<string>,
   shareId?: () => Promise<string | ShareDescriptor | ShareUrlDescriptor | void>,
   redirectUri?: string | Promise<string>,
--- a/src/Client.js
+++ b/src/Client.js
@@ -13,7 +13,7 @@ import Files from "./endpoints/Files";
 import Layers from "./endpoints/Layers";
 import Memberships from "./endpoints/Memberships";
 import Notifications from "./endpoints/Notifications";
-import Oauth from "./endpoints/Oauth";
+import OAuth from "./endpoints/OAuth";
 import Organizations from "./endpoints/Organizations";
 import Pages from "./endpoints/Pages";
 import Previews from "./endpoints/Previews";
@@ -41,7 +41,7 @@ export default class Client {
   layers: Layers;
   memberships: Memberships;
   notifications: Notifications;
-  oauth: Oauth;
+  oauth: OAuth;
   organizations: Organizations;
   pages: Pages;
   previews: Previews;
@@ -82,7 +82,7 @@ export default class Client {
     this.layers = new Layers(this, options);
     this.memberships = new Memberships(this, options);
     this.notifications = new Notifications(this, options);
-    this.oauth = new Oauth(this, options);
+    this.oauth = new OAuth(this, options);
     this.organizations = new Organizations(this, options);
     this.pages = new Pages(this, options);
     this.previews = new Previews(this, options);
--- a/src/Client.js
+++ b/src/Client.js
@@ -27,6 +27,7 @@ import Webhooks from "./endpoints/Webhooks";
 import type { CommandOptions, AnalyticsCallback } from "./types";
 
 export default class Client {
+  options: CommandOptions;
   activities: Activities;
   assets: Assets;
   branches: Branches;
@@ -56,7 +57,7 @@ export default class Client {
   _analyticsCallback: ?AnalyticsCallback;
 
   constructor(options: $Shape<CommandOptions> = {}) {
-    options = {
+    this.options = {
       accessToken: process.env.ABSTRACT_TOKEN,
       apiUrl: "https://api.goabstract.com",
       objectUrl: "https://objects.goabstract.com",
@@ -67,32 +68,37 @@ export default class Client {
       ...options
     };
 
-    this._analyticsCallback = options.analyticsCallback;
-    this.activities = new Activities(this, options);
-    this.assets = new Assets(this, options);
-    this.branches = new Branches(this, options);
-    this.changesets = new Changesets(this, options);
-    this.collectionLayers = new CollectionLayers(this, options);
-    this.collections = new Collections(this, options);
-    this.comments = new Comments(this, options);
-    this.commits = new Commits(this, options);
-    this.data = new Data(this, options);
-    this.descriptors = new Descriptors(this, options);
-    this.files = new Files(this, options);
-    this.layers = new Layers(this, options);
-    this.memberships = new Memberships(this, options);
-    this.notifications = new Notifications(this, options);
-    this.oauth = new OAuth(this, options);
-    this.organizations = new Organizations(this, options);
-    this.pages = new Pages(this, options);
-    this.previews = new Previews(this, options);
-    this.projects = new Projects(this, options);
-    this.reviewRequests = new ReviewRequests(this, options);
-    this.sections = new Sections(this, options);
-    this.shares = new Shares(this, options);
-    this.stars = new Stars(this, options);
-    this.users = new Users(this, options);
-    this.webhooks = new Webhooks(this, options);
+    this._analyticsCallback = this.options.analyticsCallback;
+    this.activities = new Activities(this, this.options);
+    this.assets = new Assets(this, this.options);
+    this.branches = new Branches(this, this.options);
+    this.changesets = new Changesets(this, this.options);
+    this.collectionLayers = new CollectionLayers(this, this.options);
+    this.collections = new Collections(this, this.options);
+    this.comments = new Comments(this, this.options);
+    this.commits = new Commits(this, this.options);
+    this.data = new Data(this, this.options);
+    this.descriptors = new Descriptors(this, this.options);
+    this.files = new Files(this, this.options);
+    this.layers = new Layers(this, this.options);
+    this.memberships = new Memberships(this, this.options);
+    this.notifications = new Notifications(this, this.options);
+    this.oauth = new OAuth(this, this.options);
+    this.organizations = new Organizations(this, this.options);
+    this.pages = new Pages(this, this.options);
+    this.previews = new Previews(this, this.options);
+    this.projects = new Projects(this, this.options);
+    this.reviewRequests = new ReviewRequests(this, this.options);
+    this.sections = new Sections(this, this.options);
+    this.shares = new Shares(this, this.options);
+    this.stars = new Stars(this, this.options);
+    this.users = new Users(this, this.options);
+    this.webhooks = new Webhooks(this, this.options);
+  }
+
+  setToken(accessToken: string) {
+    this.options.accessToken = accessToken;
+    return this.options;
   }
 
   unwrap(value: any) {
--- a/src/Client.js
+++ b/src/Client.js
@@ -98,7 +98,6 @@ export default class Client {
 
   setToken(accessToken: string) {
     this.options.accessToken = accessToken;
-    return this.options;
   }
 
   unwrap(value: any) {
--- a/src/Client.js
+++ b/src/Client.js
@@ -69,31 +69,31 @@ export default class Client {
     };
 
     this._analyticsCallback = this.options.analyticsCallback;
-    this.activities = new Activities(this, this.options);
-    this.assets = new Assets(this, this.options);
-    this.branches = new Branches(this, this.options);
-    this.changesets = new Changesets(this, this.options);
-    this.collectionLayers = new CollectionLayers(this, this.options);
-    this.collections = new Collections(this, this.options);
-    this.comments = new Comments(this, this.options);
-    this.commits = new Commits(this, this.options);
-    this.data = new Data(this, this.options);
-    this.descriptors = new Descriptors(this, this.options);
-    this.files = new Files(this, this.options);
-    this.layers = new Layers(this, this.options);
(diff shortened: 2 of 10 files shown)


src/endpoints/Oauth.js
```diff
@@ -0,0 +1,55 @@
...
+
+    const body = new URLSearchParams();
+
+    body.append("client_id", clientId);
+    body.append("client_secret", clientSecret);
+    body.append("code", authorizationCode);
+    body.append("grant_type", "authorization_code");
+    body.append("redirect_uri", redirectUri);
+
+    return this.configureRequest<Promise<TokenResponseData>>("info", {
```
amccloud, line 24:
```suggestion
    return this.configureRequest<Promise<TokenResponseData>>("getToken", {
```

src/endpoints/Oauth.js
```diff
@@ -0,0 +1,55 @@
...
+        }
+      }
+    });
+  }
+
+  setToken(accessToken: string) {
+    return new Client({
+      ...this.options,
+      accessToken
+    });
```
amccloud, line 53: @berezovskycom I wish I noticed all your code at once. I would have suggested keeping https://github.com/goabstract/abstract-sdk/pull/289 what you had originally vs making it match the spec

At the time of my suggestion I hadn't realized you created this oauth endpoint for `getToken` which does end up using `this.apiRequest` from `Endpoint`. I think you made the right choice and It would also be a consistent developer experience to have `getToken`, `setToken`, and `getAuthroizeUrl` together - like what you originally had 🙃

Endpoint options will not necessarily always match the client options so copying endpoint options to create a new client may not be not ideal

@tommoor now that we're here not sure which direction you want to go. Compared to spec, `setToken` here  is returning a new instance of Client, which I think could be preferred to mutating the original client options?

The api call above requiring this to be an endpoint to begin with may also be simple enough for calling fetch from Client?

src/endpoints/Oauth.js
```diff
@@ -0,0 +1,55 @@
...
+          `auth/tokens`,
+          {
+            method: "POST",
+            headers: {
+              "Content-Type": "application/x-www-form-urlencoded"
+            },
+            body
+          },
+          {
+            customHostname: "https://auth.goabstract.com"
```
amccloud, line 36: @tommoor still use goabstract.com or do you know if we can switch to abstract.com?

src/endpoints/Oauth.js
```diff
@@ -0,0 +1,55 @@
...
+import Endpoint from "../endpoints/Endpoint";
+import Client from "../Client";
+
+export default class OAuth extends Endpoint {
+  name = "oauth";
+
+  getToken(descriptor: AuthorizeToken) {
+    const clientId = descriptor.clientId || this.options.clientId;
+    const clientSecret = descriptor.clientSecret || this.options.clientSecret;
+    let redirectUri = descriptor.redirectUri || this.options.redirectUri;
```
amccloud, line 13: Can you also update this to an input type like in #289?

src/errors.js
```diff
@@ -48,6 +48,14 @@ export class EndpointUndefinedError extends BaseError {
...
 }
 
+export class OAuthCredentialsUndefinedError extends BaseError {
+  constructor() {
+    super(
+      "Client credentials are missing. Please doublecheck clientId, redirectUri and state"
+    );
+  }
+}
+
```
amccloud, line 58: Not used?

src/endpoints/Endpoint.js
```diff
@@ -109,7 +109,7 @@ export default class Endpoint {
     const { customHostname, raw, onProgress } = apiOptions;
     const hostname = customHostname || (await this.options.apiUrl);
 
-    fetchOptions.body = fetchOptions.body && JSON.stringify(fetchOptions.body);
+    // fetchOptions.body = fetchOptions.body && JSON.stringify(fetchOptions.body);
```
amccloud, line 112: 👀

src/endpoints/Oauth.js
berezovskycom: Hi Andrew!
Thanks for reviewing the code.
I think it would be better if to keep both requests at once, so I will add the `getAuthorizeUrl` here :)

+1 for mutating the original client options, I'm not really sure what is the best way to accomplish that.
If you do something like `this.options = { ...this.options, accessToken }`, `this.options` will be changed only in OAuth endpoint. Do you have any ideas?

tommoor:
Looks like a bunch of unrelated work from master is showing up in the PR

src/endpoints/OAuth.js
```diff
@@ -0,0 +1,75 @@
...
+    let redirectUri = input.redirectUri || this.options.redirectUri;
+    const { authorizationCode } = input;
+
+    const body = new URLSearchParams();
+
+    // $FlowFixMe
+    body.append("client_id", clientId);
+    // $FlowFixMe
+    body.append("client_secret", clientSecret);
+    // $FlowFixMe
```
amccloud, line 26: We typically try to avoid type suppressions unless it is absolutely necessary. If flow is warning here it's likely you have an unaccounted for state here

You can guard appending things to body by either providing a default 

```js
body.append("redirect_uri", window.location.href);
```

or returning early before you pass empty values to the body:

```js
if (!clientId || !clientSecret) {
  throw new Error('OAuthOnAuthorizeToken required');
}
```

src/types.js
```diff
@@ -1672,3 +1674,25 @@ export type ReviewRequest = {
...
   statusChangedAt: string
 };
+
+export type OAuthAuthorizeInput = {
+  clientId: string,
+  redirectUri: string,
+  state: string
+};
+
+export type OAuthOnAuthorizeToken = {
```
amccloud, line 1684: Since this is the input for getToken

```suggestion
export type OAuthTokenInput = {
```

`OAuthOnAuthorizeToken` seems like what `authorizationCode` is and that's only part of the input

amccloud (commented):
Thanks for consolidating the PRs 🙏 

I think once you get the commits in this PR cleaned up you should be all set for review

src/endpoints/Endpoint.js
berezovskycom: Yup, saw that. OAuth post request requires `"application/x-www-form-urlencoded"`, not JSON. Shall we add optional param to check if `JSON.stringify()` will not get used?

amccloud: Extra option or introspecting the headers to see if we're sending content-type: application/x-www-form-urlencoded

berezovskycom: Great idea!

src/endpoints/Oauth.js
tommoor: I don't know, best keep it safe.

tommoor:
Please move out of draft and re-ping when you're ready for final review.

src/endpoints/Oauth.js
tommoor: In agreement that the existing client should be mutated and returned – `Endpoint` has a `this.client` available so you should be able to mutate that and return it.

berezovskycom: Hi @tommoor @amccloud! Thanks for comments - added new commit where `client.options` gets mutated. Works as expected. Converting pr from draft to a pr and waiting for comments & approval :)

amccloud (commented):
Looks great 👍 

Before merging can we start to stub out documentation for this feature? Would be nice to cover what this PR includes

abstract-sdk.d.ts
```diff
@@ -1855,8 +1855,10 @@ type CommandOptions = {
   apiUrl: string | Promise<string>,
   analyticsCallback: AnalyticsCallback,
   assetUrl: string | Promise<string>,
+  clientSecret?: string | Promise<string>,
```
tommoor, line 1858:
```suggestion
  clientSecret?: string,
```

abstract-sdk.d.ts
```diff
@@ -1855,8 +1855,10 @@ type CommandOptions = {
...
   previewUrl: string | Promise<string>,
   shareId?: () => Promise<string | ShareDescriptor | ShareUrlDescriptor | void>,
+  redirectUri?: string | Promise<string>,
```
tommoor, line 1861:
```suggestion
  redirectUri?: string,
```

We're not supporting Promises here

src/Client.js
```diff
@@ -65,31 +68,37 @@ export default class Client {
...
-    this.pages = new Pages(this, options);
-    this.previews = new Previews(this, options);
-    this.projects = new Projects(this, options);
-    this.reviewRequests = new ReviewRequests(this, options);
-    this.sections = new Sections(this, options);
-    this.shares = new Shares(this, options);
-    this.stars = new Stars(this, options);
-    this.users = new Users(this, options);
-    this.webhooks = new Webhooks(this, options);
+    this._analyticsCallback = this.options.analyticsCallback;
```
tommoor, line 71: I think we should go ahead and just change `Endpoint` to only accept one argument, `this` – we can read `this.options` in the constructor and avoid passing it in essentially twice here.

src/endpoints/Endpoint.js
```diff
@@ -109,7 +109,15 @@ export default class Endpoint {
...
-    fetchOptions.body = fetchOptions.body && JSON.stringify(fetchOptions.body);
+    if (
+      fetchOptions.headers &&
+      fetchOptions.headers["Content-Type"] !==
+        "application/x-www-form-urlencoded"
+    ) {
+      fetchOptions.body =
+        fetchOptions.body && JSON.stringify(fetchOptions.body);
+    }
+
```
tommoor, line 120:
```suggestion
    if (
      !fetchOptions.headers || fetchOptions.headers["Content-Type"] !==
        "application/x-www-form-urlencoded"
    ) {
      fetchOptions.body =
        fetchOptions.body && JSON.stringify(fetchOptions.body);
    }

```

Suggesting this change to keep the same functionality if `headers` is present, but `Content-Type` isn't.

src/endpoints/OAuth.js
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  generateAuthorizeUrl(input: OAuthAuthorizeInput): string {
+    const clientId = input.clientId || this.options.clientId;
+    const state = input.state;
+    const redirectUri = input.redirectUri || this.options.redirectUri;
+
+    if (!clientId || !state || !redirectUri) {
+      throw new BaseError(
+        "Client credentials are missing. Please doublecheck clientId, redirectUri and state"
```
tommoor, line 60:
```suggestion
        "Client credentials are missing. Please double check clientId, redirectUri and state"
```

src/endpoints/OAuth.js
```diff
@@ -0,0 +1,68 @@
...
+      }
+    });
+  }
+
+  generateAuthorizeUrl(input: OAuthAuthorizeInput): string {
+    const clientId = input.clientId || this.options.clientId;
+    const state = input.state;
+    const redirectUri = input.redirectUri || this.options.redirectUri;
+
+    if (!clientId || !state || !redirectUri) {
```
tommoor, line 58: Is `state` strictly required?

src/endpoints/OAuth.js
```diff
@@ -0,0 +1,68 @@
...
+      throw new BaseError(
+        "Client credentials are missing. Please doublecheck clientId, redirectUri and state"
+      );
+    }
+
+    return `https://app.abstract.com/signin/auth/authorize?client_id=${clientId}&redirect_uri=${encodeURIComponent(
+      redirectUri
+    )}&response_type=code&scope=all&state=${state}`;
```
tommoor, line 66: Can we allow `input.scope` here to future proof and just set it to "all" as the default?

src/types.js
```diff
@@ -1672,3 +1675,25 @@ export type ReviewRequest = {
...
+  clientSecret: string,
+  clientId: string,
+  authorizationCode: string
+};
+
+export type TokenResponseData = {
+  access_token: string,
+  client_id: string,
+  created_at: string,
+  id: string,
```
tommoor, line 1696: What is this `id`?

src/Client.js
```diff
@@ -65,31 +68,37 @@ export default class Client {
...
+    this.sections = new Sections(this, this.options);
+    this.shares = new Shares(this, this.options);
+    this.stars = new Stars(this, this.options);
+    this.users = new Users(this, this.options);
+    this.webhooks = new Webhooks(this, this.options);
+  }
+
+  setToken(accessToken: string) {
+    this.options.accessToken = accessToken;
+    return this.options;
```
tommoor, line 101: Did you have a reason for returning options here in mind? Might be confusing, could be better to return `this` to allow for chaining or nothing at all…

tommoor (changes requested):
Almost there, I hope to get chance to spin up a sample app to really test before final approval.

src/endpoints/OAuth.js
berezovskycom: Nope. Will fix that 👍

src/endpoints/OAuth.js
berezovskycom: Good!

src/types.js
berezovskycom: According to back end, it's not used anywhere right now and not much needed for client. Looks like an id in database.

src/Client.js
berezovskycom: Yes, 100% agree. Returning nothing lgtm. Will fix that in new commit.

src/Client.js
berezovskycom: Done! I left Endpoint options param as optional, because tests were behaving weird.

src/endpoints/OAuth.js
berezovskycom: Done!

src/endpoints/OAuth.js
berezovskycom: Pushed.

src/Client.js
berezovskycom: Fixed and edited tests.

src/endpoints/Endpoint.js
```diff
@@ -37,9 +37,9 @@ export default class Endpoint {
   client: Client;
   options: CommandOptions;
 
-  constructor(client: Client, options: CommandOptions) {
+  constructor(client: Client, options?: CommandOptions) {
     this.client = client;
-    this.options = options;
+    this.options = options ? options : this.client.options;
```
berezovskycom, line 42: Not really sure if that's the best idea, but test are feeling fine.

berezovskycom:
Me & @Geek-1001 are preparing docs for OAuth, we will let you know when it gets finished!

src/endpoints/OAuth.js
```diff
@@ -54,15 +54,16 @@ export default class OAuth extends Endpoint {
     const clientId = input.clientId || this.options.clientId;
     const state = input.state;
     const redirectUri = input.redirectUri || this.options.redirectUri;
+    const scope = "all";
```
tommoor, line 57:
```suggestion
    const scope = input.scope || "all";
```

Kinda meant this

tommoor (approved):

PR #293, opened 2020-07-14
chore: add UserDescriptor to branches.list

timc1:
While working on [Abstract Home](https://goabstract.atlassian.net/browse/PLATFORM-1601)'s `My Active Branches`, I needed to grab all the branches from a specific user.

This PR adds a `UserDescriptor` as an option to the `branches.list` descriptor, so we can pass `userId` to the query params and get all branches for that user.

### DIFF
--- a/docs/abstract-api.md
+++ b/docs/abstract-api.md
@@ -226,7 +226,7 @@ abstract.branches.list({
 });
 ```
 
-List all branches for a user in a project
+List all active branches for a user in a project
 
 ```js
 abstract.branches.list({
--- a/docs/abstract-api.md
+++ b/docs/abstract-api.md
@@ -6,21 +6,20 @@ title: Client
 [cli-icon]: https://img.shields.io/badge/CLI-lightgrey.svg
 [api-icon]: https://img.shields.io/badge/API-blue.svg
 
-
 ## Activities
 
 An activity represents a designated type of event within a project. These events can be specific to the project itself, or they can be specific to a collection, a branch, a commit, or a review within the project.
 
 ### The activity object
 
-| Property      | Type                | Description                                                                             |
-|---------------|---------------------|-----------------------------------------------------------------------------------------|
-| `branchId`    | `string`            | UUID of the branch that this activity was triggered on                                  |
-| `createdAt`   | `string`            | Timestamp at which the activity was triggered                                           |
-| `id`          | `string`            | UUID identifier of the activity                                                         |
-| `payload`     | `Object`            | Object containing information specific to this type of activity                         |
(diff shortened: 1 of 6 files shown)


src/endpoints/Branches.js
```diff
@@ -58,17 +59,35 @@ export default class Branches extends Endpoint {
...
+            headers
+          });
+        }
 
-        const response = await this.apiRequest(requestUrl, { headers });
         return wrap(response.data.branches, response);
       },
 
       cli: async () => {
-        if (!descriptor) {
```
amccloud, line 71: If cli doesn't support fetching branches for a user, we'll want to add a case to warn for DX

Edit: nvm I realize this case is what I suggested, inverted

src/endpoints/Branches.js
```diff
@@ -58,17 +59,35 @@ export default class Branches extends Endpoint {
...
+        } else if (descriptor && descriptor.projectId) {
+          response = await this.apiRequest(
+            `projects/${descriptor.projectId}/branches/?${query}`,
+            { headers }
+          );
+        } else {
+          response = await this.apiRequest(`branches/?${query}`, {
+            headers
+          });
+        }
```
tommoor, line 84:
```suggestion
        if (descriptor && descriptor.userId) {
          query = querystring.stringify({
            ...queryOptions,
            userId: descriptor.userId
          });
        }

        if (descriptor && descriptor.projectId) {
          response = await this.apiRequest(
            `projects/${descriptor.projectId}/branches/?${query}`,
            { headers }
          );
        } else {
          response = await this.apiRequest(`branches/?${query}`, {
            headers
          });
        }
```

Bonus: This lets you search for a users branches within a project?

src/endpoints/Branches.js
```diff
@@ -58,17 +59,32 @@ export default class Branches extends Endpoint {
...
-          : `branches/?${query}`;
+        const queryOptions = { limit, offset, filter, search };
+        let query = querystring.stringify(queryOptions);
+        let response = null;
+
+        if (descriptor && descriptor.userId) {
+          query = querystring.stringify({
+            ...queryOptions,
+            userId: descriptor.userId
+          });
```
amccloud, line 70: I get the intention here and it'll work until you're using types

`list({ projectId, userId })` is not currently allowed so `userId` can't be present in the condition below. So `query` is only shared in the else branch but not for the projects branch

If we want to allow filtering branches for a user in a project you'll also want:

`descriptor?: ProjectDescriptor | ProjectMembershipDescriptor | UserDescriptor`

where `type ProjectMembershipDescriptor = {| projectId: string, userId: string |}`

It looks like that type exist but might be [mislabeled and mixed with `OrganizationProjectDescriptor`](https://sdk.goabstract.com/docs/abstract-api/#organizationprojectdescriptor) which doesn't appear to be used

timc1:
@amccloud thanks for the thorough review! I went ahead and updated the type definition, docs, as well as the descriptor mismatch.

docs/abstract-api.md
```diff
@@ -226,6 +226,25 @@ abstract.branches.list({
 });
 ```
 
+List all branches for a user in a project
```
tommoor, line 229:
```suggestion
List all active branches for a user in a project
```

src/endpoints/Branches.js
```diff
@@ -58,17 +59,32 @@ export default class Branches extends Endpoint {
...
-        const query = querystring.stringify({ limit, offset, filter, search });
-        const requestUrl = descriptor
-          ? `projects/${descriptor.projectId}/branches/?${query}`
-          : `branches/?${query}`;
+        const queryOptions = { limit, offset, filter, search };
+        let query = querystring.stringify(queryOptions);
+        let response = null;
+
+        if (descriptor && descriptor.userId) {
+          query = querystring.stringify({
```
tommoor, line 67: This is fine and I'm just nit'ing but you could avoid running stringify twice and neaten this up a little by just adding `userId` to the queryOptions inside of this block and then run  `querystring.stringify` afterwards?

timc1: Ty https://github.com/goabstract/abstract-sdk/pull/293/commits/7c445f02b54fcd018656e7038e41e20a20e27dc8

PR #299, opened 2020-07-24
fix: Incorrect enum for project visibility

tommoor:
feat: Add 'type' to project type (this remains purposefully undocumented for now)

### DIFF
--- a/src/types.js
+++ b/src/types.js
@@ -489,7 +489,7 @@ export type Project = {
   pushedAt: string,
   isNew: boolean,
   repoCreatedAt: string,
-  type: "cloud" | "",
+  type: "cloud" | void,
   visibility: "organization" | "specific",
   sizeInBytes: number,
   assetAutoGeneration: AssetAutoGenerationOption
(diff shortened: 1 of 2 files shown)


src/types.js
```diff
@@ -501,7 +502,8 @@ export type NewProject = {
   color?: string,
   sectionId?: string,
   createdAt?: string,
-  visibility?: "organization" | "project",
+  type?: "cloud" | "",
```
amccloud, line 505: Is `""` type expected? Maybe `void`? I think for flow you can omit this as `undefined` is already covered by `type`

tommoor: @dtakahas you know what `type` will come back as for default projects?

dtakahas: In the JSON reponse, it will be `"type": null`

tommoor: Okay, I believe this is the correct incantation now – it's optional to pass to `POST /projects`, but if passed must be `"cloud"`.

In return values from the API it will always be `"cloud"` or `null`

PR #301, opened 2020-07-27
fix: Add ValidationError class

### LINKED ISSUE #279: Do not throw MultiError unless necessary
The sdk is configured to throw MultiError [if the number of caught errors matches the number of transports configured](https://github.com/goabstract/abstract-sdk/blob/master/src/endpoints/Endpoint.js#L93). This means 1 transport with 1 error will throw a MultiError. This behavior is confusing as MultiError implies multiples errors or the possibility for multiple errors

The unecessary MultiError makes the test suite almost unworkable as any expected exceptions are masked

Screenshot 2020-06-04 15 32 34

vs

Screenshot 2020-06-04 15 38 55

I took a quick pass at fixing this but the behavior around MultiError almost always throwing is ingrained in more than a few test that I can update right now. I created this ticket In the interest of getting this fixed as I think it greatly affects someone's ability to contribute to our test and the sdk overall

cc: @berezovskycom as I know you also ran into this when you worked with test

tommoor:
Validation errors are one of the most common error types, yet we weren't giving it it's own class/handler.

Also no longer throws `MultiError` when only a single transport is used. 

closes #279

### DIFF
--- a/src/endpoints/Endpoint.js
+++ b/src/endpoints/Endpoint.js
@@ -90,7 +90,10 @@ export default class Endpoint {
         }
       }
 
-      if (Object.keys(errors).length === transportMode.length) {
+      if (
+        transportMode.length > 1 &&
+        Object.keys(errors).length === transportMode.length
+      ) {
         throw new MultiError(errors);
       }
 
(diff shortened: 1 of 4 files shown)


tommoor:
review please

PR #302, opened 2020-07-31
fix: Broken error handling

tommoor:
Error handling was totally broken by this PR: https://github.com/goabstract/abstract-sdk/pull/301 – and subsequently in the last beta release.

If a single transport is chosen and a single error happens the error handler is never called, instead you get an empty response in the success handler.

See comments for more details.

### DIFF
--- a/src/endpoints/Endpoint.js
+++ b/src/endpoints/Endpoint.js
@@ -90,10 +90,11 @@ export default class Endpoint {
         }
       }
 
-      if (
-        transportMode.length > 1 &&
-        Object.keys(errors).length === transportMode.length
-      ) {
+      const totalErrors = Object.keys(errors).length;
+      if (transportMode.length === totalErrors) {
+        if (totalErrors === 1) {
+          throw Object.values(errors)[0];
+        }
         throw new MultiError(errors);
       }
 
--- a/tests/Client.test.js
+++ b/tests/Client.test.js
@@ -19,26 +19,50 @@ describe("Client", () => {
   test("no transports specified", async () => {
     expect.assertions(1);
 
+    let error;
     try {
       await API_CLIENT.organizations.list({
         transportMode: []
       });
-    } catch (error) {
-      expect(error).toBeInstanceOf(EndpointUndefinedError);
+    } catch (e) {
+      error = e;
     }
+    expect(error).toBeInstanceOf(EndpointUndefinedError);
   });
 
   test("endpoint undefined in all transports", async () => {
     expect.assertions(2);
 
+    let error;
     try {
       await API_CLIENT.organizations.list({
         transportMode: ["api", "cli"]
       });
-    } catch (error) {
-      expect(error).toBeInstanceOf(MultiError);
-      expect(error.errors.cli).toBeInstanceOf(EndpointUndefinedError);
+    } catch (e) {
+      error = e;
     }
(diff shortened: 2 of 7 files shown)


src/endpoints/Endpoint.js
```diff
@@ -90,10 +90,11 @@ export default class Endpoint {
         }
       }
 
-      if (
-        transportMode.length > 1 &&
```
tommoor, line 94: 🍖  Errors were only thrown if both transports failed.

tests/Client.test.js
```diff
@@ -19,26 +19,50 @@ describe("Client", () => {
...
     try {
       await API_CLIENT.organizations.list({
         transportMode: []
       });
-    } catch (error) {
-      expect(error).toBeInstanceOf(EndpointUndefinedError);
+    } catch (e) {
+      error = e;
     }
+    expect(error).toBeInstanceOf(EndpointUndefinedError);
```
tommoor, line 30: 🍖  The problem with the error handling above was not caught in tests because our expectations were only checked if an error is thrown. If no error at all is thrown the test would silently pass.

I've updated all of these tests to use one of the patterns to achieve this for async methods – move `expect` out of the catch block.

tests/Client.test.js
```diff
@@ -19,26 +19,50 @@ describe("Client", () => {
...
-      expect(error).toBeInstanceOf(MultiError);
-      expect(error.errors.cli).toBeInstanceOf(EndpointUndefinedError);
+    } catch (e) {
+      error = e;
     }
+    expect(error).toBeInstanceOf(MultiError);
+    expect(error && error.errors.cli).toBeInstanceOf(EndpointUndefinedError);
+  });
+
+  test("individual transport throws", async () => {
```
tommoor, line 48: 🍖  New test for this specific case of a single transport and a single error.

src/endpoints/Endpoint.js
```diff
@@ -90,10 +90,11 @@ export default class Endpoint {
...
-        Object.keys(errors).length === transportMode.length
-      ) {
+      const totalErrors = Object.keys(errors).length;
+      if (transportMode.length === totalErrors) {
```
amccloud, line 94: This guard seems like it could be removed

tests/Client.test.js
```diff
@@ -19,26 +19,50 @@ describe("Client", () => {
...
         transportMode: ["api", "cli"]
       });
-    } catch (error) {
-      expect(error).toBeInstanceOf(MultiError);
-      expect(error.errors.cli).toBeInstanceOf(EndpointUndefinedError);
+    } catch (e) {
+      error = e;
     }
+    expect(error).toBeInstanceOf(MultiError);
+    expect(error && error.errors.cli).toBeInstanceOf(EndpointUndefinedError);
```
amccloud, line 45: Could also call `done()` after the last expect in catch and if that doesn't get called the test will throw

src/endpoints/Endpoint.js
tommoor: It can't, we only want to throw if all transports that were requested failed, if one fails and one succeeds then no throw.

tests/Client.test.js
tommoor: Because the test is async I don't think it's ideal to mix that style with the callback resolution style.

PR #309, opened 2020-09-15
feat: adds update method to Branch endpoint

enriquenov:
To make web and desktop actions work seamlessly as it relates to updating branches ([ABS-2988](https://goabstract.atlassian.net/browse/ABS-2988)), we figured it was a good move to plumb an `update` method to our SDK now that we have the server-side api endpoint available.

This change will allow us to move the `BranchUpdateRequest` to core since we'll be able to perform this request from the SDK.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,20 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## [8.0.0-beta.37](https://github.com/goabstract/abstract-sdk/compare/v8.0.0-beta.36...v8.0.0-beta.37) (2020-09-17)
+
+
+### Features
+
+* adds branches.update info to sdk docs ([eb4852b](https://github.com/goabstract/abstract-sdk/commit/eb4852b9d319c52ffcee8a0af3da3f91e5c93cc9))
+* adds tests for branches.update method ([b1ee544](https://github.com/goabstract/abstract-sdk/commit/b1ee544313423a653a033e2e70d5d5d43b6345aa))
+* adds update method to Branch endpoint ([1eef2f0](https://github.com/goabstract/abstract-sdk/commit/1eef2f07c638cfef9016b919e7c5a7a082a27376))
+
+
+### Bug Fixes
+
+* Broken wordmark image in README ([#306](https://github.com/goabstract/abstract-sdk/issues/306)) ([3507f65](https://github.com/goabstract/abstract-sdk/commit/3507f65dc556c4842de78477a542bad379a36ac1))
+
 ## [8.0.0-beta.36](https://github.com/goabstract/abstract-sdk/compare/v8.0.0-beta.35...v8.0.0-beta.36) (2020-07-31)
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,7 @@ All notable changes to this project will be documented in this file. See [standa
 
 ### Features
 
-* adds branches.update info to sdk docs ([eb4852b](https://github.com/goabstract/abstract-sdk/commit/eb4852b9d319c52ffcee8a0af3da3f91e5c93cc9))
-* adds tests for branches.update method ([b1ee544](https://github.com/goabstract/abstract-sdk/commit/b1ee544313423a653a033e2e70d5d5d43b6345aa))
-* adds update method to Branch endpoint ([1eef2f0](https://github.com/goabstract/abstract-sdk/commit/1eef2f07c638cfef9016b919e7c5a7a082a27376))
+* adds update method to Branch endpoint ([1eef2f0](https://github.com/goabstract/abstract-sdk/commit/1eef2f07c638cfef9016b919e7c5a7a082a27376)), adds respective tests ([b1ee544](https://github.com/goabstract/abstract-sdk/commit/b1ee544313423a653a033e2e70d5d5d43b6345aa)), and updates sdk docs ([eb4852b](https://github.com/goabstract/abstract-sdk/commit/eb4852b9d319c52ffcee8a0af3da3f91e5c93cc9))
 
 
 ### Bug Fixes
--- a/docs/abstract-api.md
+++ b/docs/abstract-api.md
@@ -307,7 +307,7 @@ abstract.branches.mergeState({
 
 ![CLI][cli-icon] ![API][api-icon]
 
-`branches.update(BranchDescriptor, BranchUpdateOptions, RequestOptions): Promise<Branch>`
+`branches.update(BranchDescriptor, BranchUpdateOptions): Promise<Branch>`
 
 Update an existing branch
 
--- a/docs/abstract-api.md
+++ b/docs/abstract-api.md
@@ -307,7 +307,7 @@ abstract.branches.mergeState({
 
 ![CLI][cli-icon] ![API][api-icon]
 
-`branches.update(BranchDescriptor, UpdatedBranch, RequestOptions): Promise<Branch>`
+`branches.update(BranchDescriptor, BranchInput, RequestOptions): Promise<Branch>`
 
 Update an existing branch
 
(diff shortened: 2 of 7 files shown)


src/types.js
```diff
@@ -129,6 +129,14 @@ export type RawProgressOptions = {
   onProgress?: ProgressCallback
 };
 
+export type BranchUpdateOptions = {
+  ...RequestOptions,
+  name?: string,
+  description?: string,
+  status?: string,
+  canUseAbstractd?: boolean
+};
```
enriquenov, line 138: we technically don't display a progress bar when making any of these three branch updates, so I decided not to pass the `onMessage` callback

src/endpoints/Branches.js
```diff
@@ -139,4 +140,63 @@ export default class Branches extends Endpoint {
...
+          args.push(`--status=${status}`);
+        }
+
+        if (description) {
+          args.push(`--description=${description}`);
+        }
+
+        if (canUseAbstractd) {
+          args.push(`--abstractd`);
+        }
```
enriquenov, line 192: we add the `abstractd` arg if the `canUseAbstractd` flag is truthy (which we initialize to `false`).

enriquenov:
Also, I'm a bit confused about this CircleCI failed test…

Screen Shot 2020-09-14 at 11 12 17 PM

amccloud (commented):
@enriquenov looking solid 👍 

The test failure is because the sdk enforces 100% coverage (a bit much imo) and there is currently no test for this `update` method. While you're in the process of adding test, can you also add some quick docs to abstract-api.md?

Example test: https://github.com/goabstract/abstract-sdk/blob/master/tests/endpoints/Collections.test.js#L187-L215
Example docs: https://developer.abstract.com/docs/abstract-api/#update-a-collection

enriquenov:
@amccloud thank you! and yeah, I figured that jest global coverage wasn't met, so I'm working on the specs right now 👍

docs/abstract-api.md
```diff
@@ -303,6 +303,26 @@ abstract.branches.mergeState({
 
 > Note: The API and CLI [transports](/docs/transports) behave differently for merge state. The CLI transport ignores `options.parentId`, and _only_ returns one of the three possible merge states (no other fields are included). The API transport includes a value for each field of `BranchMergeState`, and only returns statuses `CLEAN` or `NEEDS_UPDATE`.
 
+### Update a branch
+
+![CLI][cli-icon] ![API][api-icon]
+
+`branches.update(BranchDescriptor, BranchUpdateOptions): Promise<Branch>`
```
enriquenov, line 310: `RequestOptions` are included in `BranchUpdateOptions`

tests/endpoints/Branches.test.js
```diff
@@ -295,4 +295,124 @@ describe("branches", () => {
...
+        }
+      );
+
+      expect(response).toEqual({
+        branchId: "branch-id",
+        projectId: "project-id"
+      });
+    });
+
+    test("cli - with no attributes", async () => {
```
enriquenov, line 396: added a cli request test with no attributes… it was either that, or suppressing the optional checks with a bunch of `/* istanbul ignore next */` comments 🤷‍♂️

tests/endpoints/Branches.test.js
```diff
@@ -295,4 +295,124 @@ describe("branches", () => {
...
+
+  describe("update", () => {
+    test("api - with new name attribute", async () => {
+      mockAPI(
+        "/projects/project-id/branches/branch-id",
+        {
+          data: {
+            branchId: "branch-id",
+            projectId: "project-id"
+          }
```
amccloud, line 307: Should this be a `Branch`?

enriquenov: @amccloud that's what I thought initially, but then I ended up matching this `collections.update` method you suggested as an example test 😅 

https://github.com/goabstract/abstract-sdk/blob/562f7795d797dc0c0b43b30feb1cf9d161bcbeee/tests/endpoints/Collections.test.js#L187-L198

enriquenov: it seems like we don't have fake entities factories here, and we just create data that's needed for the request, which in most of the cases/examples I've seen, it evaluates to the required descriptors.

amccloud (approved):

tests/endpoints/Branches.test.js
amccloud: instanbul wants coverage for the optional cases for each value. While empty object is not a valid case in practice, the undefined behavior around empty updates allow for this test to satisfy the coverage requirement

docs/abstract-api.md
```diff
@@ -303,6 +303,26 @@ abstract.branches.mergeState({
...
+Update an existing branch
+
+```js
+abstract.branches.update({
+  projectId: "616daa90-1736-11e8-b8b0-8d1fec7aef78",
+  branchId: "master"
+}, {
+  name: "New name",
+  description: "New description",
+  status: "development"
```
ossim, line 321: nit: express that these inputs are optional?

ossim (approved):

docs/abstract-api.md
enriquenov: @ossim for updating a collection, we pass only the name option as an example even though we could also pass `description?: string, published?: boolean`: https://sdk.goabstract.com/docs/abstract-api/#update-a-collection

I want to keep things as consistent as possible with our doc, so are you suggesting to pass less options like the above case and have the user imply that not all of them are required? Because I don't see a case like this in the docs were we explicitly state that said props are optional ¯\_(ツ)_/¯

ossim: My priority when it comes to docs would be absolute consistency with the reality of how the APIs can be used. I would say that the Collection example is not great because the spec references a `RequestOptions` object that is never clearly defined - we list the properties of a Collection Object, but don't specify which of these are editable via RequestOptions. For example, would a user be able to update the CreatedAt value for a Collection?

I don't have a good answer for this, and the best change would be systematic. It's something I'd expect precedent already exist in other SDK docs though.

enriquenov:
@amccloud before I cut a new release (unless I shouldn't?), can you please review the last changes I made here:

- Split `BranchUpdateOptions` into `UpdatedBranch` and `RequestOptions. Then injected `user` param into `RequestOptions`, and lastly passed required user params to `cli` request in eed942b.
- Updated `branches.update` tests in b0b858a.

docs/abstract-api.md
amccloud: I agree that collection.update's `CollectionUpdated` is an odd one-off type. I think the recent oauth stuff added to the sdk did a good job at this. They introduced a "input" type concept. So for both these cases, you'd have `CollectionInput` and `BranchInput` types.

CHANGELOG.md
```diff
@@ -2,6 +2,20 @@
...
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## [8.0.0-beta.37](https://github.com/goabstract/abstract-sdk/compare/v8.0.0-beta.36...v8.0.0-beta.37) (2020-09-17)
+
+
+### Features
+
+* adds branches.update info to sdk docs ([eb4852b](https://github.com/goabstract/abstract-sdk/commit/eb4852b9d319c52ffcee8a0af3da3f91e5c93cc9))
+* adds tests for branches.update method ([b1ee544](https://github.com/goabstract/abstract-sdk/commit/b1ee544313423a653a033e2e70d5d5d43b6345aa))
+* adds update method to Branch endpoint ([1eef2f0](https://github.com/goabstract/abstract-sdk/commit/1eef2f07c638cfef9016b919e7c5a7a082a27376))
```
amccloud, line 12: nit: could edit this to make this a single entry

amccloud (approved):
Looks great! Before you merge and release, can you add this new update method to [the typescript definition](https://github.com/goabstract/abstract-sdk/blob/master/abstract-sdk.d.ts#L150)?

PR #318, opened 2020-11-14
fix: Add missing invariant to deps

### LINKED ISSUE #316: Missing `invariant` dependency
In https://github.com/goabstract/abstract-sdk/pull/309, the `invariant` module was imported into `src/endpoints/Branches.js`. Unfortunately, this is not declared as a dependency within `package.json`.

https://github.com/goabstract/abstract-sdk/blob/7ffcc9a54aea5d0b1c536674f2e16cca3b1f9770/src/endpoints/Branches.js#L2

This breaks usage of `abstract-sdk` unless `invariant` happens to be included as a transitive dependency of another installed package.

tommoor:
closes #316

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,6 +72,7 @@
     "core-js": "^3.6.5",
     "cross-fetch": "^3.0.1",
     "debug": "^4.0.1",
+    "invariant": "^2.2.4",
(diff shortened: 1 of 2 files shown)


enriquenov (approved):
whoops

PR #329, opened 2021-04-02
fix: increase file export timeout for api transport

rickharris:
Ran into an issue today where a file export was failing because it was taking ~45 seconds and for some reason there's a timeout imposed client-side for the `api` transport. I think we should be ok to increase this timeout or remove it altogether. I've increased it to 5 minutes in hopes to unblock a customer.


tommoor (approved):
Agreed this is a more sensible default for this operation. Maybe we can add a `timeout` override option to the method in the future

amccloud (approved):
I’m also not sure why to timeout client side. I’d vote to remove long term and let network be the source of truth. Hard to keep track of layers of timeouts 😅

PR #338, opened 2021-05-17
fix: Optional proxy agent param to fetch

### LINKED ISSUE #328: Unable to make API requests through proxy
Currently there's no way to configure a proxy for API requests.

I would suggest allowing users to optionally pass a [http-proxy-agent](https://www.npmjs.com/package/https-proxy-agent) when creating a new Abstract Client

```
const Abstract = require('abstract-sdk');
const HttpsProxyAgent = require('https-proxy-agent');

const proxyAgent = new HttpsProxyAgent('http://my-https-proxy:8080');
const client = new Abstract.Client({
  accessToken: process.env.ABSTRACT_KEY,
  transportMode: ['api'],
  proxyAgent: proxyAgent
});
```

Add optional proxyAgent to Client and update the agent to the [fetch options](https://github.com/goabstract/abstract-sdk/blob/master/src/endpoints/Endpoint.js#L118) 

```
fetchOptions.agent = Client.proxyAgent;
``` 

Solution based on [this blog post](https://www.scrapingbee.com/blog/proxy-node-fetch/)

tommoor:
I have tested locally with `mitmproxy` and it seems work great. I did think about using `node-proxy-fetch` instead to auto-detect proxy config but this suggestion from the issue I think is the most flexible and allows us to keep that configuration out of the actual module itself.

```javascript
const HttpsProxyAgent = require("https-proxy-agent");
const Abstract = require("abstract-sdk");

// change to point to your proxy or use env variable if available
const proxyAgent = new HttpsProxyAgent("http://127.0.0.1:8080");

const client = new Abstract.Client({
  proxyAgent,
  accessToken,
  ...etc
});
```

closes #328

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+### [8.1.2](https://github.com/goabstract/abstract-sdk/compare/v8.1.2-0...v8.1.2) (2021-05-21)
+
 ### [8.1.2-0](https://github.com/goabstract/abstract-sdk/compare/v8.1.1...v8.1.2-0) (2021-05-17)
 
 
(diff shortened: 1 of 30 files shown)


tommoor:
Waiting on confirmation from a user that this works for them too, if all is good I think just documentation of the option is needed to be mergable.

amccloud (approved):

